PR #10, opened 2017-01-30
Added null input argument handling

HasAndries:
I've added an implementation for null input argument handling on operations. Is this far off the mark or something missing? This is related to #9

PR #17, opened 2017-07-17
Add null argument parsing (Improved)

AndyBan:
Replaces https://github.com/graphql-dotnet/parser/pull/10

Required by https://github.com/graphql-dotnet/graphql-dotnet/pull/374

### DIFF
--- a/src/GraphQLParser/AST/Enums.cs
+++ b/src/GraphQLParser/AST/Enums.cs
@@ -25,6 +25,7 @@
         NamedType,
         ListType,
         NonNullType,
+        NullValue,
         SchemaDefinition,
         OperationTypeDefinition,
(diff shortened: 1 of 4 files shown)


src/GraphQLParser/AST/Enums.cs
```diff
@@ -25,6 +25,7 @@ public enum ASTNodeKind
         NamedType,
         ListType,
         NonNullType,
+        Null,
```
joemcbride: I think the only suggestion I have is for this to be `NullValue` vs. just `Null`.

HasAndries:
Thanks @AndyBan for this, I've been meaning to rewrite my changes in a way you've done, good job.

PR #20, opened 2018-09-18
AST printer

mkmarek:


Shane32:
I think this should code should write values to a TextWriter - no concatenation.  Ideally it should support both synchronous and asynchronous writing in case the target is a socket or filestream.  Conversion to a string can be handled by a StringWriter (which is backed by a StringBuilder).

sungam3r:
@Shane32 Do you want to work on this?

Shane32:
I think we should remove the AST model tree from GraphQL.NET, and rewrite the library to simply use the classes provided within this project.  Optionally, we could move the AST classes to a separate project that provides the models only (e.g. GraphQL.Language.Abstractions), and both projects could rely on that abstraction library of models.  It would save a lot of cpu time/memory that currently just is wasted copying information.  Then I think we may be able to remove the AST printer from there, and add it here, which seems to be a more logical place for it (being both I/O for query to string translation).

If you would agree that it is our goal, then I could be talked into working on it.  It wouldn't be hard.  But I don't have any need for two AST printers, let alone one.

Shane32:
Actually the SchemaPrinter might use a Schema, not AST.  So these printers may be different.  I'm not sure as I've never used a SchemaPrinter or AST printer.  Perhaps, however, they could share code.  Not sure without digging deeper.

sungam3r:
Let's postpone this for the next major version (not "current next").

sungam3r:
@mkmarek @Shane32 I close this in favor of #149.

PR #24, opened 2019-03-08
GraphQLSyntaxErrorException instead of ArgumentOutOfRangeException 

### LINKED ISSUE #21: Parser.Parse throws ArgumentOutOfRangeException
**Parser.Parse** throws an **ArgumentOutOfRangeException** when parsing the file from the [attached archive]. You can run the following code to reproduce it (the **path** variable should contain the path to the extracted file):

```csharp
var text = File.ReadAllText(path);
var parser = new Parser(new Lexer());
parser.Parse(new Source(text));
```

Found via [SharpFuzz].

[attached archive]: https://github.com/graphql-dotnet/parser/files/2707107/ArgumentOutOfRangeException.zip
[SharpFuzz]: https://github.com/Metalnem/sharpfuzz

sungam3r:
fixes #21

### DIFF
--- a/src/GraphQLParser.Tests/ParserTests.cs
+++ b/src/GraphQLParser.Tests/ParserTests.cs
@@ -5,11 +5,19 @@ namespace GraphQLParser.Tests
     using System.Linq;
     using GraphQLParser;
     using GraphQLParser.AST;
+    using GraphQLParser.Exceptions;
     using Xunit;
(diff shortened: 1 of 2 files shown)


joemcbride:
Looks like this has conflicts with the other PR I just merged.  Mind fixing those and I can get it merged?

PR #23, opened 2019-03-08
No allocations on GraphQLSyntaxErrorException

sungam3r:
Now the test case returns control instantly. There are no more delays for allocation of garbage strings.

``` ini

BenchmarkDotNet=v0.10.14, OS=Windows 10.0.17134
Intel Core i5-6200U CPU 2.30GHz (Skylake), 1 CPU, 4 logical and 2 physical cores
Frequency=2343747 Hz, Resolution=426.6672 ns, Timer=TSC
  [Host]     : .NET Framework 4.7.1 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3324.0
  DefaultJob : .NET Framework 4.7.1 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3324.0


```
|          Method |        Mean |     Error |     StdDev |    Gen 0 |    Gen 1 |    Gen 2 |  Allocated |
|---------------- |------------:|----------:|-----------:|---------:|---------:|---------:|-----------:|
| ParseBinaryFile | 3,537.87 us | 70.489 us | 201.110 us | 812.5000 | 750.0000 | 730.4688 | 3312.19 KB |

PR #26, opened 2019-03-20
Add initial comment support [resurrected]

sungam3r:
This is a duplicate of #16 with resolved test issues

### DIFF
--- a/src/GraphQLParser.Tests/LexerTests.cs
+++ b/src/GraphQLParser.Tests/LexerTests.cs
@@ -1,10 +1,13 @@
 ﻿namespace GraphQLParser.Tests
 {
     using GraphQLParser;
+    using System;
     using Xunit;
 
     public class LexerTests
     {
+        private static readonly string NL = Environment.NewLine;
+
         [Fact]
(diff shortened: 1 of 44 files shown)


joemcbride:
Would you mind making the `\r\n` thing cross-platform?  Use `Environment.NewLine`.  That way these tests can still be ran on macOS.  (I do most of my dev on macOS)

sungam3r:
Thanks!

Now, as I understand it, I need a new version of this library to use it in GraphQL package to bring comment support?

 Also, will you have time to look at #23 and #24 ?

joemcbride:
Yes will need a new version.  I can get those other PRs merged and publish a new nuget.

sungam3r:
Probably first it will be a prerelease nuget (for checking any unforeseen errors) which I can use in prerelease GraphQL or immediately release version? I do not really understand a good release strategy for dependent stacks of libraries.

joemcbride:
@sungam3r  I pushed `3.1.0-preview-39` of `GraphQL-Parser` to nuget.

PR #27, opened 2019-03-21
Netstandard2.0 migration

### LINKED ISSUE #25: netstandard2.0 only migration?
Recently there was a similar migration in core GraphQL repository: https://github.com/graphql-dotnet/graphql-dotnet/pull/969

Just wondering if we can do here as well. Does it make sense to continue to support net45 and netstandard1.1 TFMs ?

sungam3r:
Fixes #25


global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "1.0.4"
+    "version": "2.2.104"
```
sungam3r: I think it is possible to delete this file?

sungam3r:
Does it look good to you?

PR #36, opened 2019-07-03
Read comment for EnumValueDefinition

### LINKED ISSUE #35: Сomments are shuffled in the schema
Unfortunately, we forgot to read the comment in one of the `ParserContext` methods - `ParseEnumValueDefinition`. Because of this, the comments stack fills up and the following items take previous comments from it. As a result, the scheme turns into a mess. It was fun to watch it in our production apps. At first we thought it was our mistake ).

sungam3r:
fixes #35

also checks remaining comments in the stack

### DIFF
--- a/src/GraphQLParser.Tests/GraphQLParser.Tests.csproj
+++ b/src/GraphQLParser.Tests/GraphQLParser.Tests.csproj
@@ -10,6 +10,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Shouldly" Version="3.0.2" />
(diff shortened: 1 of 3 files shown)


sungam3r:
Wow, very responsive, thanks! Will there be a new version (4.1.1, I think) on nuget.org?

joemcbride:
Yes there will, would you mind doing that version bump?  These are all of the files that need it:  https://github.com/graphql-dotnet/parser/commit/d52e433f7f598da9ff0312ef3b5ac815114e4dbf

PR #39, opened 2019-09-02
Add api approval test

sungam3r:


sungam3r:
Same error as graphql-dotnet/graphql-dotnet#1247

sungam3r:
@joemcbride I don't know JS and for certain made changes in assembly scripts in a stupid way, but CI works.

PR #41, opened 2019-10-24
Fix error with leading pipe on directive locations and union members

sungam3r:


sungam3r:
@joemcbride Could you please push 4.1.2 to nuget.org?

joemcbride:
Published.

PR #42, opened 2019-11-16
try c# 8 features

sungam3r:
Migrate to C# 8


src/GraphQLParser/Lexer.cs
```diff
@@ -6,10 +6,8 @@ public class Lexer : ILexer
 
         public Token Lex(ISource source, int start)
         {
-            using (var context = new LexerContext(source, start))
-            {
-                return context.GetToken();
-            }
+            using var context = new LexerContext(source, start);
```
SimonCropp: some people dont like this syntax. but i think it really cleans up the code. so 👍 from me. i have moved all my repos over to it

sungam3r: Honestly, I don't like this syntax, but I think it is acceptable in such small methods.

PR #43, opened 2019-12-03
Move INamedNode to GraphQLTypeDefinition

sungam3r:
This is in order to be able to write `GraphQLTypeDefinition.Name`.

### DIFF
--- a/src/GraphQLParser/ParserContext.cs
+++ b/src/GraphQLParser/ParserContext.cs
@@ -893,6 +893,7 @@
             return new GraphQLTypeExtensionDefinition
             {
                 Comment = comment,
+                Name = definition.Name,
                 Definition = definition,
                 Location = GetLocation(start)
             };
--- a/src/GraphQLParser/ParserContext.cs
+++ b/src/GraphQLParser/ParserContext.cs
@@ -272,7 +272,7 @@
 
             if (Peek(TokenKind.NAME))
             {
-                ASTNode definition = null;
+                ASTNode definition;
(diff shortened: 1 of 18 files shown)


src/GraphQLParser/ParserContext.cs
```diff
@@ -636,28 +636,23 @@ private GraphQLName ParseName()
 
         private ASTNode ParseNamedDefinition()
         {
-            switch (currentToken.Value)
-            {
-                // Note: subscription is an experimental non-spec addition.
```
sungam3r: no more experimental non-spec addition

src/GraphQLParser/ParserContext.cs
```diff
@@ -636,28 +636,23 @@ private GraphQLName ParseName()
...
-                case "query":
-                case "mutation":
-                case "subscription":
-                    return ParseOperationDefinition();
-
-                case "fragment": return ParseFragmentDefinition();
-
-                // Note: the Type System IDL is an experimental non-spec addition.
```
sungam3r: no more experimental non-spec addition

src/GraphQLParser/ParserContext.cs
```diff
@@ -898,6 +893,7 @@ private GraphQLTypeExtensionDefinition ParseTypeExtensionDefinition()
             return new GraphQLTypeExtensionDefinition
             {
                 Comment = comment,
+                Name = definition.Name,
```
sungam3r: It was the only type inherited from `GraphQLTypeDefinition` that did not have a `Name`.

PR #50, opened 2019-12-10
[WIP] memory allocation optimizations

sungam3r:
- LexerContext, Token, GraphQLLocation -> readonly structs
- InvariantCulture compare
- new TokenKind - Unknown

This is a continuation of optimizations with allocations on the heap. In the test scenario, I managed to achieve an almost twofold reduction in allocations:

``` ini

BenchmarkDotNet=v0.11.4, OS=Windows 10.0.17134.1130 (1803/April2018Update/Redstone4)
Intel Core i5-6200U CPU 2.30GHz (Skylake), 1 CPU, 4 logical and 2 physical cores
Frequency=2343750 Hz, Resolution=426.6667 ns, Timer=TSC
.NET Core SDK=3.0.100
  [Host]     : .NET Core 3.0.0 (CoreCLR 4.700.19.46205, CoreFX 4.700.19.46214), 32bit RyuJIT
  DefaultJob : .NET Core 3.0.0 (CoreCLR 4.700.19.46205, CoreFX 4.700.19.46214), 32bit RyuJIT


```

Before optimizations. Token/GraphQLLocation/LexerContext are classes:

| Method |     Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------- |---------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
|  Parse | 3.696 us | 0.0736 us | 0.0876 us |      1.3046 |           - |           - |                2 KB |

After optimizations.Token/GraphQLLocation/LexerContext are structs:

| Method |     Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------- |---------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
|  Parse | 3.011 us | 0.0852 us | 0.1248 us |      0.7019 |           - |           - |             1.08 KB |

Unfortunately, changing the type of an object from `class` to `struct` is a backward incompatible change, so we will have to upgrade the major version. Otherwise, the design has not changed - I just needed to add a new type of token - `Unknown` because now you can not compare returned token to `null`.


src/GraphQLParser/TokenKind.cs
```diff
@@ -0,0 +1,26 @@
...
+        BRACKET_R = 11,
+        BRACE_L = 12,
+        PIPE = 13,
+        BRACE_R = 14,
+        NAME = 15,
+        INT = 16,
+        FLOAT = 17,
+        STRING = 18,
+        COMMENT = 19,
+        Unknown = 20
```
sungam3r: new enum member

sungam3r:
In fact, the bulk of the allocations load falls on the `string.Substring` method when setting `Token.Value`. We can **significantly** speed up parsing by switching to using `ReadOnlySpan<char>`. This will require `netstandard2.1` and package multitargeting + some conditional compilation directives. If the caller does not need the string value of the token, then we will not do unnecessary work. Of course, there is still a big question about the convenience of using this approach.

sungam3r:
@joemcbride  Just clarifying details from my previous comment. This approach involves postponing the calculation of the string as much as possible. Ideally, a `string` object should not be created at all. The token stores the index of the beginning and end of its position in the source text and a link to the entire source text. This is enough to get the value of the token when it needed (just reading the range of characters from start to end). Thus, we can avoid slicing the source string into many substrings in the heap. Of course, this will require the caller (actually many of them up to the stack) to handle the token accordingly. For example, we have to remove/change the `string Value`/`string Text` properties in `ASTNodes`. And above this layer is another AST layer in the graphql-dotnet project.

In general, this will be a cascade of backward incompatible destructive changes. Ultimately, it may turn out that at the next level of the call stack, we still have to materialize the string on the heap. The whole idea seems difficult to implement and maintain, although the new `System.Text.Json` parser uses just such an approach.

I just brought my thoughts here. It seems to me that at the parser level it’s worth stopping with optimizations and switching focus to higher-level things. There we can win in performance by orders of magnitude more. In the end, we should focus on the part of the code that generates the most memory traffic in the application. For a graphql server, the parser is not the main memory traffic generator - see graphql-dotnet/graphql-dotnet#1147 and graphql-dotnet/graphql-dotnet#1456

src/GraphQLParser.Benchmarks/Program.cs
```diff
@@ -1,17 +1,23 @@
...
         {
-//            var bench = new LexerBenchmark();
-//            bench.LexKitchenSink();
-            BenchmarkRunner.Run<LexerBenchmark>();
-            Console.WriteLine("===DONE===");
-            Console.ReadLine();
+            BenchmarkRunner.Run<ParserBenchmark>();
+        }
+
+        private static void Main1()
```
sungam3r: This is for memory profiler payload

sungam3r:
Closed in favor of #53

PR #70, opened 2020-04-10
Add support for descriptions to all AST nodes according to the draft specification

### LINKED ISSUE #33: Support for descriptions
So 4.0 added support for comments but it seems that comments should be ignored by the parser. Descriptions are what the spec uses for documentation. More here https://github.com/graphql/graphql-js/pull/927 and here https://graphql.github.io/graphql-spec/June2018/#sec-Descriptions

almazik:
Fixes #33 
Description parsing added to the following AST Nodes (matches the current draft specification)::

- GraphQLSchemaDefinition
- GraphQLScalarTypeDefinition
- GraphQLObjectTypeDefinition
- GraphQLFieldDefinition
- GraphQLInputValueDefinition
- GraphQLInterfaceTypeDefinition
- GraphQLUnionTypeDefinition
- GraphQLEnumTypeDefinition
- GraphQLEnumValueDefinition
- GraphQLInputObjectTypeDefinition
- GraphQLDirectiveDefinition

This pull request doesn't add the [Block Strings](https://spec.graphql.org/draft/#sec-String-Value.Block-Strings) descriptions as that should be done on lex parsing level

### DIFF
--- a/src/GraphQLParser.ApiTests/ApiApprovalTests.Public_Api_Should_Not_Change_Inadvertently.approved.txt
+++ b/src/GraphQLParser.ApiTests/ApiApprovalTests.Public_Api_Should_Not_Change_Inadvertently.approved.txt
@@ -261,7 +261,7 @@ namespace GraphQLParser.AST
     public class GraphQLSchemaDefinition : GraphQLParser.AST.ASTNode, GraphQLParser.AST.IHasDirectivesNode, GraphQLParser.AST.IHaveDescription
     {
         public GraphQLSchemaDefinition() { }
-        public GraphQLParser.AST.GraphQLDescription Description { get; set; }
+        public GraphQLParser.AST.GraphQLDescription? Description { get; set; }
         public System.Collections.Generic.List<GraphQLParser.AST.GraphQLDirective>? Directives { get; set; }
         public override GraphQLParser.AST.ASTNodeKind Kind { get; }
         public System.Collections.Generic.List<GraphQLParser.AST.GraphQLOperationTypeDefinition>? OperationTypes { get; set; }
--- a/src/GraphQLParser.ApiTests/ApiApprovalTests.Public_Api_Should_Not_Change_Inadvertently.approved.txt
+++ b/src/GraphQLParser.ApiTests/ApiApprovalTests.Public_Api_Should_Not_Change_Inadvertently.approved.txt
@@ -214,7 +214,7 @@ namespace GraphQLParser.AST
         public GraphQLParser.AST.GraphQLName? Name { get; set; }
         public GraphQLParser.AST.GraphQLValue? Value { get; set; }
     }
-    public class GraphQLObjectTypeDefinition : GraphQLParser.AST.GraphQLTypeDefinition, GraphQLParser.AST.IHasDirectivesNode, GraphQLParser.AST.IHaveDescription
+    public class GraphQLObjectTypeDefinition : GraphQLParser.AST.GraphQLTypeDefinition, GraphQLParser.AST.IHasDescriptionNode, GraphQLParser.AST.IHasDirectivesNode
     {
         public GraphQLObjectTypeDefinition() { }
         public System.Collections.Generic.List<GraphQLParser.AST.GraphQLDirective>? Directives { get; set; }
@@ -258,7 +258,7 @@ namespace GraphQLParser.AST
         public string? Value { get; set; }
         public override string? ToString() { }
     }
-    public class GraphQLSchemaDefinition : GraphQLParser.AST.ASTNode, GraphQLParser.AST.IHasDirectivesNode, GraphQLParser.AST.IHaveDescription
+    public class GraphQLSchemaDefinition : GraphQLParser.AST.ASTNode, GraphQLParser.AST.IHasDescriptionNode, GraphQLParser.AST.IHasDirectivesNode
     {
         public GraphQLSchemaDefinition() { }
         public GraphQLParser.AST.GraphQLDescription? Description { get; set; }
@@ -276,7 +276,7 @@ namespace GraphQLParser.AST
     {
         protected GraphQLType() { }
     }
-    public abstract class GraphQLTypeDefinition : GraphQLParser.AST.ASTNode, GraphQLParser.AST.IHaveDescription, GraphQLParser.AST.INamedNode
+    public abstract class GraphQLTypeDefinition : GraphQLParser.AST.ASTNode, GraphQLParser.AST.IHasDescriptionNode, GraphQLParser.AST.INamedNode
     {
         protected GraphQLTypeDefinition() { }
         public GraphQLParser.AST.GraphQLDescription? Description { get; set; }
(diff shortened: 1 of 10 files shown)


src/GraphQLParser.ApiTests/ApiApprovalTests.Public_Api_Should_Not_Change_Inadvertently.approved.txt
```diff
@@ -308,6 +317,10 @@ namespace GraphQLParser.AST
     {
         System.Collections.Generic.List<GraphQLParser.AST.GraphQLDirective>? Directives { get; set; }
     }
+    public interface IHaveDescription
```
sungam3r, line 320: Let's unify this with `IHasDirectivesNode`, so `IHasDescriptionNode`.

src/GraphQLParser.ApiTests/ApiApprovalTests.Public_Api_Should_Not_Change_Inadvertently.approved.txt
```diff
@@ -308,6 +317,10 @@ namespace GraphQLParser.AST
...
+    {
+        GraphQLParser.AST.GraphQLDescription? Description { get; set; }
```
sungam3r, line 322: As I understand it, the same elements of the schema can have descriptions, which can have comments, that is, all of them. So I propose to move this property into `ASTNode` and remove interface.

src/GraphQLParser.Tests/ParserTests.cs
```diff
@@ -530,5 +530,51 @@ public void Should_Parse_Unions(string text)
         {
             new Parser(new Lexer()).Parse(new Source(text)).ShouldNotBeNull();
         }
+
+        [Theory]
+        [InlineData("\"schema description\" schema { query : Query }", "schema description")]
+        [InlineData("\"scalar description\" scalar DateTiem", "scalar description")]
```
sungam3r, line 536:
```suggestion
        [InlineData("\"scalar description\" scalar DateTime", "scalar description")]
```

src/GraphQLParser.Tests/ParserTests.cs
```diff
@@ -530,5 +530,51 @@ public void Should_Parse_Unions(string text)
...
+        public void Should_Parse_EnumValueDefinition_Description()
+        {
+            var text = "enum Currency { \"EUR value description\" EUR \"USD value description\" USD }";
+            var document = new Parser(new Lexer()).Parse(new Source(text));
+            var enumTypeDefinition = document.Definitions.OfType<GraphQLEnumTypeDefinition>().First();
+            enumTypeDefinition.Values
+                .Select(v=>v.Description.Value)
+                .ShouldBe(new []{"EUR value description", "USD value description"});
+        }
+
```
sungam3r, line 578:
```suggestion
```

src/GraphQLParser/ParserContext.cs
```diff
@@ -240,8 +240,8 @@ private GraphQLArgument ParseArgument()
 
         private List<GraphQLInputValueDefinition>? ParseArgumentDefs()
         {
-            return Peek(TokenKind.PAREN_L)
-                ? Many(TokenKind.PAREN_L, (ref ParserContext context) => context.ParseInputValueDef(), TokenKind.PAREN_R)
+            return Peek(TokenKind.PAREN_L)
+                ? Many(TokenKind.PAREN_L, (ref ParserContext context) => context.ParseInputValueDef(), TokenKind.PAREN_R)
```
sungam3r, line 244: Hereafter, I see changes in diff view without actual changes. Could you please leave only significant changes in order to simplify and speed up the review?

src/GraphQLParser.ApiTests/ApiApprovalTests.Public_Api_Should_Not_Change_Inadvertently.approved.txt
almazik: Not all nodes can have a description according to the specification. 
See http://spec.graphql.org/draft/#sec-Document-Syntax, search for "Description". I listed all nodes which are allowed to have the description.

It is reasonable why spec designers decided to do so: there is no sense to provide documentation in the query, so only TypeSystemDefinitions should have the descriptions. Even type extensions don't have a description, I assume it's because they extend existing type and thus derive the description (documentation) from them.

src/GraphQLParser.ApiTests/ApiApprovalTests.Public_Api_Should_Not_Change_Inadvertently.approved.txt
almazik: updated

src/GraphQLParser/ParserContext.cs
almazik: Looks like Rider updated whitespaces, didn't see them on the changelist. Will update

almazik: Cleaned up the changes. It would be good to reformat the document though, it looks like the problem was with inconsistent newlines

src/GraphQLParser.ApiTests/ApiApprovalTests.Public_Api_Should_Not_Change_Inadvertently.approved.txt
sungam3r: OK

src/GraphQLParser/ParserContext.cs
sungam3r: Let's do reformatting in separate PR. I will add [.editorconfig](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/.editorconfig) some time later.

src/GraphQLParser/ParserContext.cs
```diff
@@ -627,14 +646,15 @@ private GraphQLValue ParseInt(bool isConstant)
             };
         }
 
-        private GraphQLInterfaceTypeDefinition ParseInterfaceTypeDefinition()
+        private GraphQLInterfaceTypeDefinition ParseInterfaceTypeDefinition(GraphQLDescription? description)
         {
             var comment = GetComment();
-            int start = _currentToken.Start;
+            int start = description?.Location.Start ?? _currentToken.Start;
```
sungam3r, line 652: `GraphQLDescription` node has its own location. Why do you set the start of this location to other nodes?

sungam3r: Also I think it would be better to move the `GetDescription()` inside such `ParseXXX` methods and remove the parameter. Then the code will be unified with the `GetComment()` method. How do you think?

almazik: > `GraphQLDescription` node has its own location. Why do you set the start of this location to other nodes?

Unlike comments, the description is a part of a formal definition of the parent node. So I followed the approach with other AST nodes. For example, `GraphQLObjectTypeDefinition` has a name, and so its location is wider than the corresponding `GraphQLName`'s location.

>Also I think it would be better to move the GetDescription() inside such ParseXXX methods and remove the parameter. Then the code will be unified with the GetComment() method. How do you think?

I tried to place the description parsing closer to the specific ParseXxx, but there was a problem: when we encounter a string (description), we don't know what kind of node we are parsing yet. So we have to read past the description node in order to find out the name of the entity and branch the logic to the corresponding ParseXxx method. That was the reason why I refactored `ParseNamedDefinition` into 2 methods: `ParseNamedDefinition` which parses named nodes without description (and thus immediately knows which node we are at), and `ParseNamedDocumentedDefinition` which is called when we encounter a string literal (description).

prowe:
If this PR has no blocking issues, My team would really appreciate this getting merged..

almazik:
We eventually switched to HotChocolate due to blocking issues in graphql-dotnet and no clear progress in resolving them

sungam3r:
I am looking forward to reviewing several PRs this month. Too much work, too little free time.

Shane32:
Superseded by #132

PR #71, opened 2020-06-05
Fix some source location bugs

ificator:
While developing in `graphql-dotnet` I noticed a couple of issues related to the values for node locations in the original query:
1. The indexes were plain wrong because the constructor of `Source` performs some replace operations before parsing and so the indexes are not relative to the original query.
2. The value for `End` for a lot of nodes actually includes the first token immediately following the node. For example, if I have the following:
```
fragment f1 on type {
  ...
}

fragment f2 on type {
  ...
}
```
Using the location data for the first fragment to drive a substring of the original query (ignoring issue 1) results in:
```
fragment f1 on type {
  ...
}

fragment
```

The parser already treats `\r` and `\n` as expected, so fixing the first issue appears to be as simple as not string replacing. As a bonus this avoids potentially expensive allocations for large queries.

The second issue also has a relatively simple fix, as long as my understanding of the state of the parser holds true. At the time the location indices are evaluated the `_currentToken` has progressed to the token after the node of interest, and so the value for `End` should simply be `_currentToken.Start`.

### DIFF
--- a/src/GraphQLParser/Source.cs
+++ b/src/GraphQLParser/Source.cs
@@ -1,4 +1,4 @@
-﻿namespace GraphQLParser
+namespace GraphQLParser
 {
     public class Source : ISource
     {
@@ -9,18 +9,11 @@
         public Source(string body, string name)
         {
             Name = name;
-            Body = MonetizeLineBreaks(body);
+            Body = body ?? string.Empty;
         }
 
         public string Body { get; set; }
 
         public string Name { get; set; }
-
-        private static string MonetizeLineBreaks(string input)
-        {
-            return (input ?? string.Empty)
-                .Replace("\r\n", "\n")
-                .Replace("\r", "\n");
-        }
     }
 }
--- a/src/GraphQLParser.Tests/LexerTests.cs
+++ b/src/GraphQLParser.Tests/LexerTests.cs
@@ -6,7 +6,9 @@ namespace GraphQLParser.Tests
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "IDE0022:Use expression body for methods", Justification = "Tests")]
     public class LexerTests
     {
-        private static readonly string NL = Environment.NewLine;
+        // Use a constant newline instead of Environment.NewLine to make sure the tests are
+        // consistently executed between Windows and *nix.
+        private static readonly string NL = "\n";
 
         [Fact]
         public void Lex_ATPunctuation_HasCorrectEnd()
--- a/src/GraphQLParser.Tests/ParserTests.cs
+++ b/src/GraphQLParser.Tests/ParserTests.cs
@@ -295,9 +295,39 @@ scalar JSON
         }
 
         [Fact]
+        public void Parse_Fields_Object_LocationCorrect()
+        {
+            var document = ParseGraphQLFieldsSource();
+            var query = GetSingleOperationDefinition(document);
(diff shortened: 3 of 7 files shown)


src/GraphQLParser/Source.cs
```diff
@@ -9,18 +9,11 @@ public Source(string body) : this(body, "GraphQL")
...
             Name = name;
-            Body = MonetizeLineBreaks(body);
+            Body = body ?? string.Empty;
         }
 
         public string Body { get; set; }
 
         public string Name { get; set; }
-
-        private static string MonetizeLineBreaks(string input)
```
sungam3r, line 19: Once I already wanted to remove this method but did not dare.

sungam3r:
Could you please revert all changes with line endings - `Environment.NewLine` <-> `\n`? These changes do not apply to the described problem and only complicate the code review. You can publish separate PR for them later if you want.

src/GraphQLParser/Source.cs
ificator: Yeah I debated allowing the behavior to be controlled using an optional constructor parameter but reasoned that I was overcomplicating things. I think the only time someone could potentially encounter an issue with this change is if the query contains a string parameter and an implementation is relying on the `\r\n` -> `\n` conversion the parser is performing. That seemed unlikely (but not impossible).

ificator:
> Could you please revert all changes with line endings - `Environment.NewLine` <-> `\n`? These changes do not apply to the described problem and only complicate the code review. You can publish separate PR for them later if you want.

Some of them do - after the removal of the monetization a number of these tests failed, and a couple of the fixes required the removal of the blind string replace. I could have only touched the offending test cases, but that would have left inconsistency in the test implementations that I despise :). It also seemed "bad" to have test cases that relied on the line endings of the source file so I opted to fix them all while I had the hood open. If you disagree with this I can revert the collateral and leave only the impacted tests as currently written.

sungam3r:
> At the time the location indices are evaluated the `_currentToken` has progressed to the token after the node of interest,

Right

> and so the value for `End` should simply be `_currentToken.Start`.

Not really. Actually it should be equal to `prevToken.End` since there is at least one whitespace symbol before `_currentToken`. So usage of `_currentToken.Start` as the location end will give lengthening of the whole token. Thanks for the idea anyway, see #78. We can continue to discuss the first problem with `MonetizeLineBreaks` after merging #78 if you don't mind to resolve conflicts.

ificator:
Thanks @sungam3r - the modified string is available in the `Source`, so fixing the busted end value is the primary concern. Once your change is in I'll merge and we can discuss further.

sungam3r:
#78 was merged.

sungam3r:
@ificator will you continue to work on this PR or open another one for `MonetizeLineBreaks`?

Shane32 (approved):

Shane32:
Removing `MonetizeLineBreaks` is important now that with GraphQL 3.0 we are retrieving the position properties directly from the parser, so the positions must relate to the input.

Shane32:
Of course, this PR needs to be updated, or replaced with another, due to #78 being merged, as @sungam3r noted.

ificator:
Hey folks, thanks for the ping - I'll clean this up and push a new iteration

ificator:
I did it anyway because I like consistency in code :)

https://github.com/graphql-dotnet/parser/pull/86

Since the combination of PR 78 and PR 86 fully cover this PR I'm going to close it.

PR #73, opened 2020-08-08
Add properties to exception

Shane32:
This PR adds 3 new properties to the `GraphQLSyntaxErrorException` class:
* Description
* Line
* Column

The `Message` property will not change.  All tests that verify the `Message` property have been updated to verify the values in these new fields.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: 5.1.2.{build}
+version: 5.1.3.{build}
 pull_requests:
   do_not_increment_build_number: true
 skip_tags: true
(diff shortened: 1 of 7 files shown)


Shane32:
See https://github.com/graphql-dotnet/graphql-dotnet/pull/1765

These added properties will allow graphql-dotnet to return location information from the parser back to the response graph.

sungam3r:
Thanks. You should also bump minor version.

sungam3r:
See https://github.com/graphql-dotnet/parser/pull/67/files

Shane32:
Sorry about that.  I've not published any nuget packages myself yet, so not used to versioning.  Our internal projects just use a date code.

PR #75, opened 2020-08-15
Verify CI checks

Shane32:
This PR should fail CI checks.  Do not merge.


Shane32:
@sungam3r If you look at the first red X, there are two tests, and one shows a green check -- but clicking the details shows the test obviously failed.  However after a code change, now both tests fail.  Something's not working correctly.

PR #78, opened 2020-08-15
Fix location errors

sungam3r:
Fixes token location bug. Thanks @ificator for idea. Relates to #71 . See #71 for more info.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: 5.2.0.{build}
+version: 5.3.0.{build}
 pull_requests:
   do_not_increment_build_number: true
 skip_tags: true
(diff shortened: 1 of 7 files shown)


sungam3r:
@Shane32 you might be interested.

ificator:
Thanks for doing this - I agree with persisting the previous token and using that for a more correct end value. I wasn't familiar enough with the code to drill in that far :).

The only other piece I wouldn't mind pulling in from my PR is the comment for `GraphQLLocation.Start` and `GraphQLLocation.End`. When I started poking around it wasn't obvious that `Start` was inclusive and `End` was exclusive and I think it'll help people in the future.

sungam3r:
OK. Done.

PR #79, opened 2020-08-15
Bump test deps

sungam3r:


src/GraphQLParser.Tests/GraphQLParser.Tests.csproj
```diff
@@ -11,11 +11,12 @@
 
   <ItemGroup>
     <PackageReference Include="Shouldly" Version="4.0.0-beta0002" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.7.0" />
```
sungam3r, line 14: Added explicitly since it was removed from `xunit.runner.visualstudio` dependencies.

PR #83, opened 2020-10-07
String.concat optimization for STRING and COMMENT tokens

sungam3r:
I came across the fact that the parser was processing some requests for a very long time and with the allocation of a huge amount (tens, hundreds of megabytes) of memory. It turned out that this happens when parsing strings (for example, in arguments represented by large objects with many properties, including nested ones), when the string contains escape sequences, i.e. `\"`. I saw that the code was doing many string concatenations, forming the final string from small chunks. The algorithm now uses a buffer. The parser processed one of my requests for more than 10 seconds, now it happens almost instantly. The memory consumption has been reduced many times over.

### DIFF
--- a/src/GraphQLParser/LexerContext.cs
+++ b/src/GraphQLParser/LexerContext.cs
@@ -1,5 +1,6 @@
 using GraphQLParser.Exceptions;
 using System;
+using System.Text;
 
 namespace GraphQLParser
 {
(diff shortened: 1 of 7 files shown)


src/GraphQLParser/LexerContext.cs
```diff
@@ -308,28 +341,6 @@ private int GetPositionAfterWhitespace(string body, int start)
...
-        {
-            if (_currentIndex + 5 > _source.Body.Length)
-            {
-                string truncatedExpression = _source.Body.Substring(_currentIndex);
-                throw new GraphQLSyntaxErrorException($"Invalid character escape sequence at EOF: \\{truncatedExpression}.", _source, _currentIndex);
-            }
-
-            string expression = _source.Body.Substring(_currentIndex, 5);
-
-            if (!OnlyHexInString(expression.Substring(1)))
```
sungam3r, line 321: no more `.Substring(1)`

sungam3r:
It can be difficult to figure out what exactly has changed. I changed the code for `ReadString` and `ReadComment`. It even got easier.

Shane32:
I didn't read through the whole thing, but the feature seems great!  I assume that all of the code runs off of a `string` (or `char` array), which means it's already been converted from UTF-8 to UTF-16, and multibyte unicode characters should pass through the parsing engine without any problems. Correct?

sungam3r:
This PR is not about encoding and multibyte characters, it is about escape sequences and string concatenations. See #84 for performance results.

Shane32:
Of course it isn't.  My comment is simply that now that you've changed how the parser works, the new code should work with multibyte characters.  Or at least not work any worse than before.  I assume all is well and you can merge this at your leisure.  It will be a nice performance increase when using string literals.

sungam3r:
> Or at least not work any worse than before.

it seems so

joemcbride:
Awesome work!

sungam3r:
Now the problematic graphql request in our project began to reach the resolver instantly even with more larger string argument - 3 MB.

Shane32:
Sounds great!!!!! 🥇

PR #86, opened 2020-10-09
Stop query Normalization

ificator:
This is derived from https://github.com/graphql-dotnet/parser/pull/71 due to it being easier to resolve conflicts from a clean slate.

Specifically:
```
1. The indexes were plain wrong because the constructor of `Source` performs some replace operations before parsing and
   so the indexes are not relative to the original query

...

The parser already treats \r and \n as expected, so fixing the first issue appears to be as simple as not string replacing.
As a bonus this avoids potentially expensive allocations for large queries.
```

### DIFF
--- a/src/GraphQLParser/Source.cs
+++ b/src/GraphQLParser/Source.cs
@@ -1,4 +1,4 @@
-﻿namespace GraphQLParser
+namespace GraphQLParser
 {
     public class Source : ISource
     {
@@ -9,18 +9,11 @@
         public Source(string body, string name)
         {
             Name = name;
-            Body = MonetizeLineBreaks(body);
+            Body = body ?? string.Empty;
         }
 
         public string Body { get; set; }
 
         public string Name { get; set; }
-
-        private static string MonetizeLineBreaks(string input)
-        {
-            return (input ?? string.Empty)
-                .Replace("\r\n", "\n")
-                .Replace("\r", "\n");
-        }
     }
 }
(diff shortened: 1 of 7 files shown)


Shane32 (approved):

Shane32:
@sungam3r This looks good to me. Want to review it also?

joemcbride (approved):

sungam3r:
Yes

src/GraphQLParser/Source.cs
```diff
@@ -9,18 +9,11 @@ public Source(string body) : this(body, "GraphQL")
         public Source(string body, string name)
         {
             Name = name;
-            Body = MonetizeLineBreaks(body);
```
sungam3r, line 12: @joemcbride  Why bother doing that?

sungam3r:
> The parser already treats \r and \n as expected, so fixing the first issue appears to be as simple as not string replacing.

Hmm. Now I looked at the code and saw that it really is. I used to think there was some reason to call the `MonetizeLineBreaks` method. Then I don't understand why it was ever needed.

> As a bonus this avoids potentially expensive allocations for large queries.

This is not a bonus, this is the main purpose of this PR. I have long regretted allocating memory at this location.

sungam3r (approved):

sungam3r:
@ificator Please bump version. Here is an example https://github.com/graphql-dotnet/parser/pull/85/files

sungam3r:
https://www.nuget.org/packages/GraphQL-Parser/5.3.3

PR #89, opened 2020-10-27
Add codecoverage, CodeQL analysis, dotnet-format

sungam3r:


sungam3r:
It works! https://coveralls.io/builds/34609689

src/Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <VersionPrefix>5.3.5-preview</VersionPrefix>
+    <VersionPrefix>5.3.4-preview</VersionPrefix>
```
sungam3r, line 4: Shoul be 4, my mistake

src/Directory.Build.targets
```diff
@@ -0,0 +1,23 @@
+<!-- This target must be imported into Directory.Build.targets -->
+<!-- Workaround. Remove once we're on 3.1.300+
+https://github.com/dotnet/sourcelink/issues/572 -->
```
sungam3r, line 3: I will delete this file with the next coverlet release. The problem is already fixed in master.

sungam3r:
@joemcbride @Shane32 PR is finally ready for review. I added new actions in test job for calculating code coverage.

Modified steps:
1. dotnet test began to produce lcov.info coverage file

New steps:
1. Convert coverage report to clover format ('Monitor coverage' works only with this format)
2. Monitor coverage adds coverage comment in PR (replace mode, only one comment)
3. Upload coverage to Coveralls, see https://coveralls.io/github/graphql-dotnet/parser
4. Upload coverage to Codecov. Now not working and disabled. I wanted to see what it looks like, but I haven’t succeeded yet. It seemed to me that coveralls has a not very intuitive UI.

Questions:
1. Do I need to move all new actions to run only on Linux?
2. Is it possible and necessary to do this in a separate job, so as not to resort to `if:` condition for steps?
3. What to do with master branch? It would be nice to do a test coverage on `master` too.
4. Should I try to get Codecov work (may need help).

After we get everything in order here, I would like to add the same steps to other repositories.

README.md
```diff
@@ -1,8 +1,8 @@
 # GraphQL Dotnet Parser
-[![AppVeyor](https://img.shields.io/appveyor/ci/graphql-dotnet-ci/parser.svg)](https://ci.appveyor.com/project/graphql-dotnet-ci/parser)
+
 [![Coverage Status](https://coveralls.io/repos/github/graphql-dotnet/parser/badge.svg?branch=master)](https://coveralls.io/github/graphql-dotnet/parser?branch=master)
```
sungam3r, line 3: Note - it still points to old builds (2017 year) from `master`, but coverage is triggered now only for `test` job.

Shane32: The `test` job will never run on `master`.  Do you intend to add code coverage generation upon merge also?  Maybe modify the `test` job so that it also runs on `push` to `master` or `develop`.

src/Directory.Build.targets
Shane32: (`master` of coverlet)

.github/workflows/test.yml
```diff
@@ -37,4 +37,45 @@ jobs:
...
+          reporttypes: 'Clover'
+          tag: '${{ github.run_number }}_${{ github.run_id }}'
+      - name: Monitor coverage
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        uses: slavcodev/coverage-monitor-action@1.2.0
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          clover_file: "Clover.xml"
+          threshold_alert: 80
+          threshold_warning: 90
```
Shane32, line 55: Can you explain what this does?  It would seem that it triggers some type of alert or failure if code coverage drops below a certain percentage?  Isn't that a bit pointless?  I mean, we really want to know the amount of coverage for the code modified in the PR, not as a global percentage.  It's not like people are going to change 10% of the codebase.  Or is that what this does?

.github/workflows/test.yml
```diff
@@ -37,4 +37,45 @@ jobs:
...
         working-directory: src
-        run: dotnet test --no-restore
+        run: dotnet test --no-restore -p:CollectCoverage=true -p:CoverletOutputFormat=lcov -p:Include=[GraphQL-Parser]* -p:CoverletOutput=../../${{ matrix.os }}.lcov.info
+      - name: Convert coverage report to clover format
+        uses: danielpalme/ReportGenerator-GitHub-Action@4.7.1
+        with:
+          reports: '${{ matrix.os }}.lcov.info'
+          targetdir: '.'
+          reporttypes: 'Clover'
+          tag: '${{ github.run_number }}_${{ github.run_id }}'
```
Shane32, line 47: How is this tag used?  Is this a required format?  Perhaps `'test_${{ github.run_number }}` would be better.  The run number will be unique for this yml file, within this repository.

Shane32 (approved):
Very nice!

src/Directory.Build.targets
sungam3r: yes

README.md
sungam3r: > The test job will never run on master.

I know.

> Do you intend to add code coverage generation upon merge also?

Yes.

> Maybe modify the test job so that it also runs on push to master or develop.

I wanted to get advice on how to do better. I understand that we can make as many files as we like. Many stages in these files will be repeated (checkout/build/test/etc). We can place all actions even in one file to "save steps", but then we get a large and complex file. I would like to make the files simpler, let there be more of them. Then each file will correspond to its own scenario. How do you think?

.github/workflows/test.yml
sungam3r: https://github.com/graphql-dotnet/parser/pull/89#issuecomment-719935986

This is a comment showing the percentage of code coverage by tests.

sungam3r: This is nothing more than supplementary information that may be useful for consideration. If the PR is such that the coverage drops significantly, then this is a reason to think.

.github/workflows/test.yml
sungam3r: So far this tag is not used in any way. ReportGenerator was needed to just convert from one format to another: lcov -> clover. If you use ReportGenerator for a straightforward purpose - to render coverate reports in html - then this tag is displayed in the report. I've been using ReportGenerator for this on my internal projects for a long time.

README.md
Shane32: I really like the way it's laid out now.  It seems pretty simple.  As I recommended, I would add a couple lines to this same `test.yml` file to make it run when PRs are merged:

```
  push:
    branches:
      - master
      - develop
```

I would not separate test coverage into a separate workflow -- basically all the steps of producing code coverage reports are necessary for running the tests in the first place.  The only thing you could do is skip running on windows. In that case, sure we can create a separate workflow file.  (Or we can add a bunch of if conditions.)

I might also add some more comments to the workflow file(s) -- but, the workflow steps are already labeled, so even that isn't necessary.  Anyone can read through the file and see exactly what the script is doing.  GitHub Action workflows are very easy to follow.

I certainly would not combine the workflow files we have.  Some of the build steps differ - for example, the nuget package location while setting up .net.  This would make the workflow file overly complex, and it would be harder for someone to understand what's going on.

I don't mind the if conditions for operating system, because the steps all DO run within the workflow -- just some only once, and others twice.  We don't have if conditions based on whether it was a 'pull_request' or 'push' trigger, for instance, which fundamentally changes what steps the workflow runs.

.github/workflows/test.yml
Shane32: Hmm ok

Shane32:
Do you think we should add "html" and "htmlsummary" outputs to the build artifacts for the test?

See: https://github.com/Shane32/ExcelLinq/pull/3/files for necessary changes

Sample output: download artifacts from here:
https://github.com/Shane32/ExcelLinq/actions/runs/340792028

I'm going to do this for my private repos instead of paying for coveralls pro.

sungam3r:
I was expecting this question. I also use ReportGenerator on my projects at work. We use TeamCity and there it is possible to display artifacts on a separate project tab, just point to some 'root' file to render (index.html). It's convenient, you don't need to download anything and the report is always in front of your eyes. The report really helps to understand what the code is not covered by the tests. In the case of GitHub Action, I'm not sure that someone will download this report as artifact. At least I won't, it's not convenient for me and I'm just lazy. For me it is much more convenient to configure systems such as coveralls or codecov to do all the work of displaying and maintaining coverage statistics. Of course, the question is how convenient these systems are and their cost. I have no experience with them.

> I'm going to do this for my private repos instead of paying for coveralls pro.

What gives pro compared to what is given for free?


As a conclusion. I was going to read general information about different code coverage web systems to compare them and evaluate their pros and cons.

Shane32:
Coveralls free isn’t free for private repos. If you’ve got a better solution for private repos, I’m all ears.

sungam3r:
I don't khow, my primary goal are public repos like GraphQL.NET, Parser. etc.

README.md
sungam3r: done

.github/workflows/test.yml
sungam3r: done

.github/workflows/test.yml
sungam3r: deleted

src/Directory.Build.targets
sungam3r: deleted

PR #101, opened 2021-01-10
string -> ROM

sungam3r:
This PR brings `System.Memory` bits into lexer and parser. Also redesigned to reduce the number of memory allocations. Lexer has stopped allocating memory on the heap at all (with rare exceptions for escaped symbols handling). The parser still allocates memory, since it builds AST, but already noticeably less. Benchmarks, however, do not show the unconditional leadership of the new version in terms of execution time. I have noticed that my results vary from run to run.

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.16299.726 (1709/FallCreatorsUpdate/Redstone3)
Intel Core i7 CPU 920 2.67GHz (Nehalem), 1 CPU, 8 logical and 4 physical cores
Frequency=2607423 Hz, Resolution=383.5204 ns, Timer=TSC
.NET Core SDK=5.0.101
  [Host]     : .NET Core 3.1.10 (CoreCLR 4.700.20.51601, CoreFX 4.700.20.51901), X64 RyuJIT
  DefaultJob : .NET Core 3.1.10 (CoreCLR 4.700.20.51601, CoreFX 4.700.20.51901), X64 RyuJIT

```

Lexer before/after
| Method |          name |           Mean |       Error |      StdDev |    Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |-------------- |---------------:|------------:|------------:|---------:|------:|------:|----------:|
|    **Lex** |       **escapes** |    **99,603.5 ns** |    **64.93 ns** |    **60.74 ns** |   **3.9063** |     **-** |     **-** |   **16384 B** |
|    **Lex** |        **github** | **4,346,489.0 ns** | **3,373.87 ns** | **3,155.92 ns** | **179.6875** |     **-** |     **-** |  **783024 B** |
|    **Lex** |          **hero** |       **659.9 ns** |     **0.77 ns** |     **0.65 ns** |   **0.0362** |     **-** |     **-** |     **152 B** |
|    **Lex** | **introspection** |    **22,313.3 ns** |    **15.03 ns** |    **13.33 ns** |   **0.8240** |     **-** |     **-** |    **3560 B** |
|    **Lex** |       **kitchen** |    **18,486.7 ns** |    **15.56 ns** |    **14.56 ns** |   **0.6409** |     **-** |     **-** |    **2736 B** |
|    **Lex** |        **params** |    **11,780.0 ns** |    **13.36 ns** |    **12.50 ns** |   **0.7629** |     **-** |     **-** |    **3248 B** |

| Method |          name |           Mean |       Error |      StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |-------------- |---------------:|------------:|------------:|-------:|------:|------:|----------:|
|    **Lex** |       **escapes** |    **89,692.2 ns** |   **122.49 ns** |   **114.58 ns** | **1.7090** |     **-** |     **-** |    **7344 B** |
|    **Lex** |        **github** | **4,406,522.9 ns** | **1,651.93 ns** | **1,289.72 ns** |      **-** |     **-** |     **-** |       **4 B** ? |
|    **Lex** |          **hero** |       **788.0 ns** |     **1.49 ns** |     **1.39 ns** |      **-** |     **-** |     **-** |         **-** |
|    **Lex** | **introspection** |    **23,013.3 ns** |    **20.99 ns** |    **19.63 ns** |      **-** |     **-** |     **-** |         **-** |
|    **Lex** |       **kitchen** |    **20,923.3 ns** |     **2.84 ns** |     **2.65 ns** |      **-** |     **-** |     **-** |         **-** |
|    **Lex** |        **params** |    **16,792.0 ns** |     **8.95 ns** |     **8.37 ns** |      **-** |     **-** |     **-** |         **-** |

Parser before/after
| Method |          name |         Mean |      Error |     StdDev |    Gen 0 |    Gen 1 | Gen 2 | Allocated |
|------- |-------------- |-------------:|-----------:|-----------:|---------:|---------:|------:|----------:|
|  **Parse** |       **escapes** |   **100.903 μs** |  **0.1188 μs** |  **0.1053 μs** |   **4.0283** |   **0.1221** |     **-** |   **17312 B** |
|  **Parse** |        **github** | **9,620.084 μs** | **12.3197 μs** | **11.5239 μs** | **390.6250** | **187.5000** |     **-** | **2492766 B** |
|  **Parse** |          **hero** |     **1.355 μs** |  **0.0011 μs** |  **0.0009 μs** |   **0.2308** |        **-** |     **-** |     **968 B** |
|  **Parse** | **introspection** |    **35.247 μs** |  **0.0318 μs** |  **0.0297 μs** |   **3.9063** |   **0.0610** |     **-** |   **16432 B** |
|  **Parse** |       **kitchen** |    **29.227 μs** |  **0.0204 μs** |  **0.0181 μs** |   **3.0823** |   **0.0305** |     **-** |   **12960 B** |
|  **Parse** |        **params** |    **18.617 μs** |  **0.0109 μs** |  **0.0097 μs** |   **2.3499** |   **0.0305** |     **-** |    **9944 B** |

| Method |          name |         Mean |      Error |     StdDev |    Gen 0 |    Gen 1 | Gen 2 | Allocated |
|------- |-------------- |-------------:|-----------:|-----------:|---------:|---------:|------:|----------:|
|  **Parse** |       **escapes** |    **83.919 μs** |  **0.0319 μs** |  **0.0266 μs** |   **1.9531** |        **-** |     **-** |    **8288 B** |
|  **Parse** |        **github** | **6,264.506 μs** | **28.3585 μs** | **25.1391 μs** | **296.8750** | **148.4375** |     **-** | **1824992 B** |
|  **Parse** |          **hero** |     **1.421 μs** |  **0.0013 μs** |  **0.0012 μs** |   **0.1965** |        **-** |     **-** |     **824 B** |
|  **Parse** | **introspection** |    **33.082 μs** |  **0.0199 μs** |  **0.0186 μs** |   **3.1738** |        **-** |     **-** |   **13512 B** |
|  **Parse** |       **kitchen** |    **29.282 μs** |  **0.0267 μs** |  **0.0236 μs** |   **2.5330** |        **-** |     **-** |   **10696 B** |
|  **Parse** |        **params** |    **22.740 μs** |  **0.0215 μs** |  **0.0191 μs** |   **1.5564** |        **-** |     **-** |    **6592 B** |

@Shane32 Please try run `LexerBenchmark` and `ParserBenchmark`.


sungam3r:
It should also be noted that this PR opens the way for further optimizations at the GraphQL.NET level. We can now reuse the same portions of the pooled memory over and over. Of course this will require changes to GraphQL.NET because it manipulates strings from the parser. Now there are no strings, only ROM. In principle, it is feasible to achieve not spawning strings at all. After that, we can evaluate the final performance

src/GraphQLParser/Token.cs
```diff
@@ -1,24 +1,27 @@
...
 
-        public int End { get; }
+        }
 
         public TokenKind Kind { get; }
 
         public int Start { get; }
 
-        public string? Value { get; }
+        public int End { get; }
```
sungam3r, line 20: I had an idea to replace `int` with `short` to reduce the size of the structure, but then long queries will stop working. It would be possible to solve this problem somehow using a generic parameter, but I did not complicate the code.

src/GraphQLParser/Location.cs
```diff
@@ -6,26 +7,20 @@ namespace GraphQLParser
...
-        public Location(string source, int position)
+        public Location(ReadOnlySpan<char> source, int position)
         {
             Line = 1;
             Column = position + 1;
 
             if (position > 0)
             {
-                var matches = _lineRegex.Matches(source);
+                var matches = _lineRegex.Matches(source.ToString()); // TODO: heap allocation
```
sungam3r, line 17: I didn’t find the method I needed to run regexp againts ROM/Span.

src/GraphQLParser/Lexer.cs
```diff
@@ -1,15 +1,18 @@
-﻿namespace GraphQLParser
+using System;
+
+namespace GraphQLParser
 {
-    public class Lexer : ILexer
+    public struct Lexer
     {
-        public ILexemeCache? Cache { get; set; }
+        public int? BufferSize { get; set; }
```
sungam3r, line 7: By the way, the lexer can still allocate memory in the case of parsing strings with escape characters.

src/GraphQLParser/GraphQLParser.csproj
```diff
@@ -4,11 +4,16 @@
...
     <AssemblyTitle>GraphQL Parser for .NET</AssemblyTitle>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
     <AssemblyName>GraphQL-Parser</AssemblyName>
     <PackageTags>GraphQL;json;api;parser</PackageTags>
+    <NoWarn Condition="'$(TargetFramework)' == 'netstandard2.1'">$(NoWarn);IDE0057</NoWarn> <!--Slice can be simplified-->
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.Memory" Version="4.5.4" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
```
sungam3r, line 14: @Shane32 Returning to our talk about `System.Memory` dependency. It can then be removed from GraphQL.NET.

src/GraphQLParser.Tests/LexerTests.cs
```diff
@@ -31,7 +31,7 @@ public void Lex_ATPunctuation_HasCorrectStart()
         public void Lex_ATPunctuation_HasCorrectValue()
         {
             var token = GetATPunctuationTokenLexer();
-            token.Value.ShouldBeNull();
+            token.Value.ToString().ShouldBe("@");
```
sungam3r, line 34: It is worth paying attention to this. I began to return the actual contents of the token, not null.

src/GraphQLParser.Tests/LexerTests.cs
```diff
@@ -353,7 +353,7 @@ public void Lex_NameTokenWithWhitespaces_HasNameKind()
         [Fact]
         public void Lex_NullInput_ReturnsEOF()
         {
-            var token = new Lexer().Lex(new Source(null));
+            var token = new Lexer().Lex((string)null);
```
sungam3r, line 356: https://github.com/dotnet/roslyn/issues/50360

Shane32:
```
*******develop********

| Method |          name |           Mean |       Error |      StdDev |    Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------- |-------------- |---------------:|------------:|------------:|---------:|-------:|------:|----------:|
|    Lex |       escapes |    49,942.7 ns |   146.38 ns |   136.93 ns |   2.5635 | 0.0610 |     - |   16385 B |
|    Lex |        github | 2,223,245.5 ns | 2,710.56 ns | 2,402.84 ns | 121.0938 |      - |     - |  783027 B |
|    Lex |          hero |       325.9 ns |     4.64 ns |     4.34 ns |   0.0238 |      - |     - |     152 B |
|    Lex | introspection |    11,655.1 ns |    20.61 ns |    18.27 ns |   0.5646 |      - |     - |    3560 B |
|    Lex |       kitchen |     9,004.7 ns |    29.17 ns |    27.29 ns |   0.4272 |      - |     - |    2736 B |
|    Lex |        params |     5,791.0 ns |    14.21 ns |    12.60 ns |   0.5112 |      - |     - |    3248 B |

| Method |          name |           Mean |        Error |       StdDev |    Gen 0 |    Gen 1 | Gen 2 | Allocated |
|------- |-------------- |---------------:|-------------:|-------------:|---------:|---------:|------:|----------:|
|  Parse |       escapes |    49,589.0 ns |     93.14 ns |     82.57 ns |   2.7466 |   0.1831 |     - |   17312 B |
|  Parse |        github | 4,808,853.2 ns | 25,318.60 ns | 23,683.03 ns | 390.6250 | 195.3125 |     - | 2492666 B |
|  Parse |          hero |       659.3 ns |      1.94 ns |      1.81 ns |   0.1535 |   0.0010 |     - |     968 B |
|  Parse | introspection |    18,018.1 ns |     58.53 ns |     54.75 ns |   2.5940 |   0.2441 |     - |   16432 B |
|  Parse |       kitchen |    13,738.3 ns |     24.74 ns |     23.15 ns |   2.0599 |   0.1678 |     - |   12960 B |
|  Parse |        params |     9,047.5 ns |     33.89 ns |     31.70 ns |   1.5717 |   0.0763 |     - |    9944 B |

**********rom***********

| Method |          name |           Mean |       Error |      StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |-------------- |---------------:|------------:|------------:|-------:|------:|------:|----------:|
|    Lex |       escapes |    41,670.3 ns |    43.24 ns |    38.34 ns | 1.1597 |     - |     - |    7344 B |
|    Lex |        github | 2,045,813.8 ns | 6,475.18 ns | 6,056.89 ns |      - |     - |     - |       2 B |
|    Lex |          hero |       351.6 ns |     0.57 ns |     0.54 ns |      - |     - |     - |         - |
|    Lex | introspection |    10,610.0 ns |    21.66 ns |    19.20 ns |      - |     - |     - |         - |
|    Lex |       kitchen |     9,431.1 ns |    21.26 ns |    17.75 ns |      - |     - |     - |         - |
|    Lex |        params |     8,484.7 ns |    13.47 ns |    11.94 ns |      - |     - |     - |         - |

| Method |          name |           Mean |       Error |      StdDev |    Gen 0 |    Gen 1 | Gen 2 | Allocated |
|------- |-------------- |---------------:|------------:|------------:|---------:|---------:|------:|----------:|
|  Parse |       escapes |    42,258.5 ns |   143.60 ns |   134.32 ns |   1.2817 |        - |     - |    8288 B |
|  Parse |        github | 3,462,496.7 ns | 8,169.69 ns | 7,242.22 ns | 289.0625 | 144.5313 |     - | 1824999 B |
|  Parse |          hero |       781.9 ns |     1.68 ns |     1.49 ns |   0.1307 |        - |     - |     824 B |
|  Parse | introspection |    18,809.6 ns |    34.68 ns |    30.74 ns |   2.1362 |   0.0916 |     - |   13512 B |
|  Parse |       kitchen |    15,064.9 ns |    39.29 ns |    36.75 ns |   1.6937 |   0.0610 |     - |   10696 B |
|  Parse |        params |    12,286.3 ns |    18.02 ns |    16.85 ns |   1.0376 |   0.0153 |     - |    6592 B |
```

sungam3r:
Thanks. I updated results in my initial comment after making Lexer and Parser static.

Shane32:
> Thanks. I updated results in my initial comment after making Lexer and Parser static.

Ditto

src/GraphQLParser/LexerContext.cs
```diff
@@ -90,8 +76,7 @@ public Token ReadNumber()
 
             if (nextCode >= 48 && nextCode <= 57)
```
Shane32, line 77:
```suggestion
            if (nextCode >= '0' && nextCode <= '9')
```
I suggest that the character representations be used rather than the value throughout the code.

src/GraphQLParser/LexerContext.cs
```diff
@@ -116,94 +101,163 @@ public Token ReadNumber()
             return isFloat ? CreateFloatToken(start) : CreateIntToken(start);
         }
 
+        public Token ReadComment()
+        {
+            int start = _currentIndex;
+            char code = NextCode();
+
+            Span<char> buffer = stackalloc char[4096];
+            StringBuilder? sb = null;
```
Shane32, line 110: Perhaps a single StringBuilder can be allocated and shared for the entire context, so using pooling is not necessary.  (Or the context can grab a stringbuilder from a pool, and release it before returning.)

src/GraphQLParser/LexerContext.cs
```diff
@@ -116,94 +101,163 @@ public Token ReadNumber()
...
+
         public Token ReadString()
         {
             int start = _currentIndex;
             char code = NextCode();
-            var sb = new StringBuilder(); // TODO: buffer allocation; it can be pooled later
 
-            while (IsNotAtTheEndOfQuery() && code != 0x000A && code != 0x000D && code != '"')
+            Span<char> buffer = stackalloc char[4096];
+            StringBuilder? sb = null;
```
Shane32, line 163: Then the same stringbuilder can be used here.

src/GraphQLParser/Location.cs
Shane32: We should eliminate all regex.  It should be extremely easy, especially for this code!  Of course this is only used for error messages, but still...

src/GraphQLParser/ParserContext.cs
```diff
@@ -689,23 +692,48 @@ private GraphQLName ParseName()
...
-                "union" => ParseUnionTypeDefinition(),
-                "enum" => ParseEnumTypeDefinition(),
-                "input" => ParseInputObjectTypeDefinition(),
-                "extend" => ParseTypeExtensionDefinition(),
-                "directive" => ParseDirectiveDefinition(),
-                _ => null
-            };
+            var s = _currentToken.Value.Span;
+
+            if (s.SequenceEqual("query".AsSpan()))
```
Shane32, line 697: Would it be faster to convert `"query".AsSpan()` into a constant or anything?  Probably not but just asking.

src/GraphQLParser/Token.cs
Shane32: I would leave it as an `int`.

Shane32:
I looked through the code once.  Is it ready to be merged or are you still working on it?  I would like to take another look.  Also, I think we should not publish this to nuget until 4.0 is ready to be released, as we may find more changes we want to make.  We can reference the development build via nuget from GraphQL.NET

src/GraphQLParser/LexerContext.cs
sungam3r: done

src/GraphQLParser/LexerContext.cs
sungam3r: It really is. This is actually a rare case when dealing with very large string literals. One can also take ROM from the pool. But this will complicate the code, I just didn’t want to make it even more complicated without an obvious need. Maybe later.

sungam3r: I'm not using `StringBuilder` here because I like it. It stores a linked list of array objects inside itself, so there can still be a lot of memory allocations. Ideally, it would be working with one large array (behind ROM) from the pool. This array needs to be returned to the pool, so I will have to resort to `IMemoryOwner` services. It will need to be passed up the call stack, stored, etc.

src/GraphQLParser/LexerContext.cs
sungam3r: see https://github.com/graphql-dotnet/parser/pull/101#discussion_r554628143

src/GraphQLParser/Location.cs
sungam3r: I know. I'm not ready to fix this code yet. In any case, the fix can be made without changing the public api in future releases. Published #102 .

src/GraphQLParser/ParserContext.cs
sungam3r: `Span` is a `ref struct` so it can be stored only on stack. Moreover these methods are almost instantaneous.

sungam3r:
I spent a lot of time trying to make the lex/parse algorithms faster. I don't see any obvious way to speed things up. I don't want to change anything else. Now I have removed almost all memory allocations associated with strings, there are a few rare cases that can be fixed later, the public api should not change. I agree that it's not worth publishing the package just yet. We need to take a working dev version of it and make a PR in GraphQL.NET.

sungam3r:
I have to admit that I hate to see that not all test cases are accelerated. But I believe that the ability to use the memory pool is much more important now. We significantly reduce the load on the garbage collector in the context of the entire application. Let the execution time decrease slightly in some simple scenarios, but the memory stability will increase. It may turn out that by using the pool, benchmarks in GraphQL.NET will start to work faster. Until we check, we will not understand.

sungam3r:
I also need to rewrite readme, but later.

sungam3r:
@joemcbride Are you OK with this direction of design?

src/GraphQLParser/LexerContext.cs
```diff
@@ -8,27 +8,25 @@ namespace GraphQLParser
     internal struct LexerContext
```
Shane32, line 8:
```suggestion
    internal ref struct LexerContext
```

src/GraphQLParser/LexerContext.cs
```diff
@@ -48,38 +46,26 @@ public Token GetToken()
...
-
-        public Token ReadComment()
+        private static bool OnlyHexInString(ReadOnlySpan<char> test)
         {
-            int start = _currentIndex;
-            char code = NextCode();
-            var sb = new StringBuilder(); // TODO: buffer allocation; it can be pooled later
-
-            while (IsNotAtTheEndOfQuery() && code != 0x000A && code != 0x000D)
+            for (int i=0; i<test.Length; ++i)
```
Shane32, line 54:
```suggestion
            for (int i=0; i < test.Length; ++i)
```

src/GraphQLParser/LexerContext.cs
```diff
@@ -402,10 +447,12 @@ private Token ReadName()
...
             {
-                _currentIndex++;
-                code = GetCode();
+                if (++_currentIndex < _source.Length)
+                    code = _source.Span[_currentIndex];
+                else
+                    break;
             }
-            while (IsNotAtTheEndOfQuery() && IsValidNameCharacter(code));
+            while (code == '_' || char.IsLetterOrDigit(code));
```
Shane32, line 455:
```suggestion
            while (code == '_' || ('a' <= code && code <= 'z') || ('A' <= code && code <= 'Z') || ('0' <= code & code <= '9'));
```
This is faster and meets the spec exactly.

src/GraphQLParser/LexerContext.cs
```diff
@@ -48,38 +46,26 @@ public Token GetToken()
             if (code == '"')
```
Shane32, line 46: A few lines above, you can change `char.IsLetter(code)` to `('a' <= code && code <= 'z') || ('A' <= code && code <= 'Z')` for faster performance.  Similar with `IsNumber`.  No need to check for unicode variations of the alphabet if the spec doesn't allow for it.

Shane32 (approved):

src/GraphQLParser/LexerContext.cs
sungam3r: I have already tried this. It's getting slower 😕

src/GraphQLParser/LexerContext.cs
sungam3r: Did you specifically write & instead of &&?

src/GraphQLParser/LexerContext.cs
sungam3r: done

sungam3r:
Finally! I was able to achieve a noticeable improvement by throw helpers to allow code inlining. I updated results in my initial comment.

sungam3r:
Also interesting https://github.com/dotnet/runtime/issues/13347

src/GraphQLParser/LexerContext.cs
Shane32: No, it was an accident.  But a single ampersand might be faster, due to less conditional jump statements and possibly less chance of branch prediction errors in the CPU pipeline.  I have not tested it so I've been just using &&

src/GraphQLParser/LexerContext.cs
Shane32: Doesn't this result in exactly the same compiled code, just placing additional restrictions on use?  Is it actually slower to add the ref in here?

src/GraphQLParser/LexerContext.cs
sungam3r: I thought about it too, but didn't use.

src/GraphQLParser/LexerContext.cs
sungam3r: I'm just saying what I tried and it turned out to be slower. I've tried different approaches. You can experiment yourself, benchmarks are already prepared.

Shane32: lol - maybe if I have time I will just because I'm curious.  But I'm very happy going forward based on your findings.  I think we just need to wrap up some of these PRs so we can keep making progress.

src/GraphQLParser/Parser.cs
```diff
@@ -1,22 +1,10 @@
...
-        }
-
-        public GraphQLDocument Parse(ISource source) => Parse(source, false);
-
-        public GraphQLDocument Parse(ISource source, bool ignoreComments)
-        {
-            var context = new ParserContext(source, _lexer, ignoreComments);
-            return context.Parse();
-        }
+        public static GraphQLDocument Parse(ReadOnlyMemory<char> source, bool ignoreComments = true) => new ParserContext(source, ignoreComments).Parse();
```
sungam3r, line 8: @Shane32 I'm sure that in v7 we should ignore comments by default to increase the parsing speed. I will write about it in the readme. Or you if you want.

src/GraphQLParser/LexerContext.cs
sungam3r: I think your latest changes with `ignorecomments` make this problem less important since default value for it is `true` in v7.

src/GraphQLParser/ParserContext.cs
```diff
@@ -592,19 +605,23 @@ private GraphQLFragmentDefinition ParseFragmentDefinition()
...
+        private void Throw_From_ParseFragmentName()
+        {
+            throw new GraphQLSyntaxErrorException($"Unexpected {_currentToken}", _source.Span, _currentToken.Start);
+        }
+
         private List<GraphQLNamedType>? ParseImplementsInterfaces()
         {
             List<GraphQLNamedType>? types = null;
-            if (_currentToken.Value?.Equals("implements") == true)
+            if (_currentToken.Value == "implements")
```
sungam3r, line 624: @Shane32 With new syntax sugar all conditions look better.

src/GraphQLParser/ROM.cs
```diff
@@ -0,0 +1,100 @@
...
+        public bool Equals(ROM other)
+        {
+            if (_memory.Length != other._memory.Length)
+                return false;
+
+            return _memory.Span.SequenceEqual(other._memory.Span);
+        }
+
+        /// <inheritdoc/>
+        public override int GetHashCode()
```
sungam3r, line 50: Taken from `string` implementation.

src/GraphQLParser/ROM.cs
```diff
@@ -0,0 +1,100 @@
+using System;
+
+namespace GraphQLParser
+{
+    /// <summary>
+    /// A wrapper around ReadOnlyMemory{char} allowing you to use simple syntax when working with it.
```
sungam3r, line 6: The main problem with `ReadOnlyMemory{char}` is its `GetHashCode` : 
```c#
 /// <summary>Returns the hash code for this <see cref="ReadOnlyMemory{T}"/></summary>
        [EditorBrowsable(EditorBrowsableState.Never)]
        public override int GetHashCode()
        {
            // We use RuntimeHelpers.GetHashCode instead of Object.GetHashCode because the hash
            // code is based on object identity and referential equality, not deep equality (as common with string).
            return (_object != null) ? HashCode.Combine(RuntimeHelpers.GetHashCode(_object), _index, _length) : 0;
        }
```

My experiments have shown that with such an implementation it is impossible to work with dictionaries, because the hash code for matching pieces of a string will still be different:

example: `abc abc`

`Slice(0, 3).GetHashCode() != Slice(4, 3).GetHashCode()`

In our case I compare **content** of memory chunks.

Shane32: You can specify a comparer for a dictionary to use which overrides the hash code and equality comparisons. https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.comparer?view=net-5.0

sungam3r: I still need ==, != and implicit operators so it is easier to keep all this stuff inside one custom struct.

src/GraphQLParser/ROM.cs
```diff
@@ -81,6 +81,8 @@ public override int GetHashCode()
 
         public static implicit operator ReadOnlyMemory<char>(ROM rom) => rom._memory;
 
+        public static implicit operator ReadOnlySpan<char>(ROM rom) => rom._memory.Span;
```
Shane32, line 84: For memory on the managed heap, does this create a copy onto the stack?  If yes, maybe it wouldn't be such a great idea to have as an implicit operator.

sungam3r: No. `ReadOnlySpan` is a "smart" struct pointing to some chunk of memory - either heap or stack. `Memory.Span` dereference does nothing bad.

src/GraphQLParser/ROM.cs
```diff
@@ -0,0 +1,102 @@
...
+        public static implicit operator ROM(char[] array) => new ReadOnlyMemory<char>(array);
+
+        public static bool operator ==(ROM rom, string s) => rom._memory.Span.SequenceEqual(s.AsSpan());
+
+        public static bool operator !=(ROM rom, string s) => !rom._memory.Span.SequenceEqual(s.AsSpan());
+
+        public static bool operator ==(string s, ROM rom) => rom == s;
+
+        public static bool operator !=(string s, ROM rom) => rom != s;
+    }
```
Shane32, line 101:
```suggestion

        public static bool operator ==(ROM rom1, ROM rom2) => rom1.Equals(rom2);

        public static bool operator !=(ROM rom1, ROM rom2) => !rom1.Equals(rom2);
    }
```

sungam3r: 🤦 I forgot about the most obvious operator.

PR #105, opened 2021-01-15
Add capability to ignore comments during parsing

Shane32:
When combined with #101 this should provide a reduction in memory requirements for queries with comments when executing queries where comments are superfluous.  This adheres to the spec, which says that comments should be treated as whitespace.

> Comments behave like white space and may appear after any token, or before a line terminator, and have no significance to the semantic meaning of a GraphQL Document.


sungam3r:
Let's resolve conflicts and merge this before #101.

src/GraphQLParser.Tests/ParserTests.cs
```diff
@@ -56,6 +56,25 @@ public void Extra_Comments_Should_Read_Correclty()
...
+        public void Comments_Can_Be_Ignored()
+        {
+            const string query = @"
+{
+    #comment
+    person
+}";
+
+            var parser = new Parser(new Lexer());
+            var document = parser.Parse(new Source(query), true);
```
sungam3r, line 69:
```suggestion
            var document = parser.Parse(new Source(query), true);
            document .UnattachedComments.ShouldBeNull();
```

src/GraphQLParser.Tests/ParserTests.cs
```diff
@@ -56,6 +56,25 @@ public void Extra_Comments_Should_Read_Correclty()
             document.UnattachedComments[2].Text.ShouldBe("comment4");
         }
 
+        [Fact]
+        public void Comments_Can_Be_Ignored()
+        {
+            const string query = @"
+{
+    #comment
+    person
```
sungam3r, line 65:
```suggestion
    person
    # comment2
```

src/GraphQLParser.Tests/ParserTests.cs
```diff
@@ -56,6 +56,27 @@ public void Extra_Comments_Should_Read_Correclty()
...
+    # comment2
+}";
+
+            var parser = new Parser(new Lexer());
+            var document = parser.Parse(new Source(query), true);
+            document .UnattachedComments.ShouldBeNull();
```
sungam3r, line 71:
```suggestion
            document.UnattachedComments.ShouldBeNull();
```

sungam3r (approved):

PR #108, opened 2021-01-20
Add xml docs

sungam3r:


src/GraphQLParser/Location.cs
```diff
@@ -2,10 +2,18 @@
 
 namespace GraphQLParser
 {
+    /// <summary>
+    /// Provides the ability to decode a linear character position into a row and column number.
```
sungam3r, line 6:
```suggestion
    /// Provides the ability to decode a linear character position into a line and column number.
```

src/GraphQLParser/AST/ASTNodeKind.cs
```diff
@@ -0,0 +1,77 @@
...
+        Document,
+        OperationDefinition,
+        VariableDefinition,
+        Variable,
+        SelectionSet,
+        Field,
+        Argument,
+        FragmentSpread,
+        InlineFragment,
+        FragmentDefinition,
```
Shane32, line 23: Applying a comment to `Name` does not automatically apply it to the other enumeration entries.

Shane32 (approved):
Looks like a good start!  Any idea how many more warnings are left?

src/GraphQLParser/AST/ASTNodeKind.cs
Shane32: (does it?)

sungam3r: Of course I added comment above only for `Name`.

sungam3r:
I'm not a native speaker. It's better to double-check what I wrote in the comments.

Shane32:
I read all the comments.  They seem fine.

PR #112, opened 2021-01-22
Remove extension methods

### LINKED ISSUE #111: develop branch: ParserExtensions methods should not be extensions
I don't think we should have extension methods called `Parse` and `Lex` on string.

Why should `"test".Parse()` return a `GraphQLDocument`?  Why shouldn't it parse to an integer, for example?  And why should `"test".Lex()` create a lexer for GraphQL documents?  Why not a lexer for json?

When you parse an integer, you call `int.Parse("234")`.  It's not just naming -- Microsoft didn't add a "ParseInt" extension method for strings.

I believe we should remove `ParseExtensions`. The static methods on `Parser` and `Lexer` are well designed.  No extension methods are needed.  If desired for tests, someone can add an extension method within their a test project to facilitate testing.  It should not be part of the library.

Shane32:
Fixes #111

### DIFF
--- a/src/GraphQLParser.Tests/Extensions.cs
+++ b/src/GraphQLParser.Tests/Extensions.cs
@@ -1,6 +1,3 @@
-using System;
-using System.Collections.Generic;
-using System.Text;
 using GraphQLParser.AST;
 
 namespace GraphQLParser.Tests
--- a/src/GraphQLParser.Tests/Extensions.cs
+++ b/src/GraphQLParser.Tests/Extensions.cs
@@ -1,0 +1,28 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using GraphQLParser.AST;
+
+namespace GraphQLParser.Tests
+{
+    internal static class ParserTestExtensions
+    {
+
+        /// <summary>
+        /// Generates token based on input text.
+        /// </summary>
+        /// <param name="source">Input data as a string.</param>
+        /// <param name="start">The index in the source at which to start searching the token.</param>
+        /// <returns></returns>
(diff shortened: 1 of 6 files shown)


src/GraphQLParser.Tests/Extensions.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using GraphQLParser.AST;
+
+namespace GraphQLParser.Tests
+{
+    internal static class ParserTestExtensions
+    {
+
```
sungam3r, line 10:
```suggestion
```

src/GraphQLParser.Tests/Extensions.cs
```diff
@@ -0,0 +1,28 @@
...
+        public static Token Lex(this string source, int start = 0) => Lexer.Lex(source, start);
+
+        /// <summary>
+        /// Generates AST based on input text.
+        /// </summary>
+        /// <param name="source">Input data as a string.</param>
+        /// <param name="options">Parser options.</param>
+        /// <returns>AST (Abstract Syntax Tree) for GraphQL document.</returns>
+        public static GraphQLDocument Parse(this string source, ParserOptions options = default) => Parser.Parse(source, options);
+
```
sungam3r, line 26:
```suggestion
```

sungam3r:
OK. Please change also readme.

sungam3r (approved):

sungam3r:
Now see if #111 closes automatically.

sungam3r:
Nope

Shane32:
Yeah, weird.

sungam3r:
https://github.blog/2013-01-22-closing-issues-via-commit-messages/

PR #114, opened 2021-02-01
net5 support for tests and Coveralls integration

sungam3r:


sungam3r:
@Shane32 See the comment above. This has not happened before.

sungam3r:
@Shane32 Since the integration with coveralls worked and now the code coverage change is visible, I removed `Monitor coverage` and `Convert` steps. I like how it works now.

Issue #107, opened 2021-01-16
Release v7


sungam3r:
1. ~Test string -> ROM in GraphQL.NET~ DONE
2. ~Update README.md~ DONE
3. ~Describe new option for ignoring comments.~ DONE in #101
4. ~XML comments~ PARTIALLY DONE
5. ~Take care of #89~ DONE

sungam3r:
@Shane32 The main problem with v7 in GraphQL.NET is that new ROM APIs destroy familiar string-based interfaces and approaches.

sungam3r:
@Shane32 Are we ready to release v7?

Shane32:
As far as I know, but let’s release it alongside v4

sungam3r:
Done in https://github.com/graphql-dotnet/parser/releases/tag/7.0.0

PR #116, opened 2021-02-01
allow empty object/input/interface/enum type definitions

csboling:
A schema definition such as:

```
type Query
```

with no `FieldsDefinition` was rejected: 

```
   GraphQLParser.Exceptions.GraphQLSyntaxErrorException : Syntax Error GraphQL (1:11) Expected {, found EOF
1: type Query
```

The GraphQL spec indicates that for [object types](https://spec.graphql.org/June2018/#ObjectTypeDefinition), [input object types](https://spec.graphql.org/June2018/#InputObjectTypeDefinition), [interfaces](https://spec.graphql.org/June2018/#InterfaceTypeDefinition), and [enums](https://spec.graphql.org/June2018/#EnumTypeDefinition) the curly-brace enclosed field set is optional, and notes that for [type extensions](https://spec.graphql.org/June2018/#ObjectTypeExtension) you may wish to extend a type just to add a directive or to indicate that it implements an interface without declaring any new fields.

The reference implementation has some test cases for these rules [here](https://github.com/graphql/graphql-js/blob/dc89262a5ecca0d48b3d5e54d5cb726762e358fd/src/language/__tests__/schema-parser-test.js#L202). Note that the reference implementation rejects type definitions which include curlies but do not declare fields, e.g. `type Query { }`, which the graphql-dotnet parser accepts. Currently this means that there is no way to write "forward declarations" of object types in a way that both the graphql-js and graphql-dotnet parsers will allow. See some discussion [here](https://github.com/graphql/graphql-spec/issues/568).

### DIFF
--- a/src/GraphQLParser.Tests/ParserTests.cs
+++ b/src/GraphQLParser.Tests/ParserTests.cs
@@ -523,5 +523,17 @@ Cat
         {
             new Parser(new Lexer()).Parse(new Source(text)).ShouldNotBeNull();
         }
+
+        [Theory]
+        [InlineData("type Query", ASTNodeKind.ObjectTypeDefinition)]
+        [InlineData("extend type Query", ASTNodeKind.TypeExtensionDefinition)]
+        [InlineData("input Empty", ASTNodeKind.InputObjectTypeDefinition)]
+        [InlineData("extend type Type implements Interface", ASTNodeKind.TypeExtensionDefinition)]
+        public void Should_Parse_Empty_Types(string text, ASTNodeKind kind)
+        {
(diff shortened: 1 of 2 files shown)


sungam3r:
Thanks. See #107. Please retarget PR on `develop`.

sungam3r:
I'll handle it myself.

sungam3r:
@csboling Thanks again. This PR is replaces by #117.

> Note that the reference implementation rejects type definitions which include curlies but do not declare fields, e.g. type Query { }, which the graphql-dotnet parser accepts.

Also fixed.

PR #126, opened 2021-04-06
Add support for '&' syntax token

sungam3r:
Fixes https://github.com/graphql-dotnet/graphql-dotnet/issues/2478

This is nonsense. I can not imagine how all this time this obvious error remained unnoticed. Parser did not support the `&` token all this time. Ping @joemcbride  and @Shane32 . It is possible that I do not understand something, but in all apparently no one GraphQL.NET client used more than one interface on the object.

### DIFF
--- a/src/GraphQLParser.ApiTests/GraphQL-Parser.approved.txt
+++ b/src/GraphQLParser.ApiTests/GraphQL-Parser.approved.txt
@@ -455,5 +455,6 @@ namespace GraphQLParser
         STRING = 18,
         COMMENT = 19,
         UNKNOWN = 20,
+        AMPERSAND = 21,
     }
 }
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Setup .NET Core 5.0 SDK
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: '5.0.x'
+          dotnet-version: '5.0.201'
           source-url: https://nuget.pkg.github.com/graphql-dotnet/index.json
         env:
           NUGET_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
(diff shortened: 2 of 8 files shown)


src/GraphQLParser.ApiTests/GraphQL-Parser.approved.txt
```diff
@@ -455,5 +455,6 @@ namespace GraphQLParser
         STRING = 18,
         COMMENT = 19,
         UNKNOWN = 20,
+        AMPERSAND = 21,
```
sungam3r, line 458: It is unpleasant that the `UNKNOWN` value ceased to be the last. To save backward compatibility...

sungam3r:
OOOPs, new SDK here!

v5.0.202
Released
    2021-04-06 


v5.0.201:
``` 
Registered Sources:
  1.  Source [Enabled]
      https://nuget.pkg.github.com/graphql-dotnet/index.json
  2.  nuget.org [Enabled]
      https://api.nuget.org/v3/index.json
  3.  Microsoft Visual Studio Offline Packages [Enabled]
      C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\
```

v5.0.202
```
Registered Sources:
  1.  Source [Enabled]
      https://nuget.pkg.github.com/graphql-dotnet/index.json
  2.  Microsoft Visual Studio Offline Packages [Enabled]
      C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\
```

Temporary downgraded to v5.0.201

joemcbride (approved):

joemcbride:
> This is nonsense. I can not imagine how all this time this obvious error remained unnoticed.

Not too common to use seems like!

Shane32:
> I can not imagine how all this time this obvious error remained unnoticed.

🤷‍♂️ I don't use interfaces at all.

> Temporary downgraded to v5.0.201

Well apparently Microsoft can't even release a new version of their SDK that runs on their own operating system.... 🤣

Even funnier, the reason they released a new version is to FIX nuget package restoring 🤣🤣🤣🤣🤣🤣

https://devblogs.microsoft.com/nuget/net-5-nuget-restore-failures-on-linux-distributions-using-nss-or-ca-certificates/

Shane32 (approved):

sungam3r:
> Interestingly, the reason they released a new version was to FIX nuget package restoring...

Yep! What is interesting I know about this bug and waited for the fix. We could not migrate to .NET 5 because of it. One corrected, the other broke ;)

PR #131, opened 2021-06-28
Add support for BlockStrings

Shane32:
As a BlockString is a form of a StringValue, no changes were made to the public API.  A BlockString is allowed anywhere a StringValue is accepted, pursuant to the current (June 2018) specification.

Considering the scope of the change, I propose that the version number be bumped to 7.2.0, even though there are no public API changes.

The new `ReadBlockString()` method first reads the string, unescaping `\"""` to `"""` and replacing line termination sequences with `\n`.  Then the string value is passed to a secondary method (a static local function) to strip leading and trailing blank lines and unindent the text if needed, all according to spec.  If the string is small enough, both of these allocations occur on the stack.  For larger sequences, heap allocation may be necessary.  A multitude of tests are added to cover the range of indenting scenarios and so on that are provided for by the spec.

### DIFF
--- a/src/GraphQLParser.Tests/Validation/LexerValidationTests.cs
+++ b/src/GraphQLParser.Tests/Validation/LexerValidationTests.cs
@@ -391,5 +391,40 @@ namespace GraphQLParser.Tests.Validation
             exception.Line.ShouldBe(1);
             exception.Column.ShouldBe(14);
         }
+
+        [Theory]
+        [InlineData("test", "test")]
+        [InlineData("te\\\"\"\"st", "te\"\"\"st")]
+        [InlineData("\ntest", "test")]
+        [InlineData("\r\ntest", "test")]
+        [InlineData(" \ntest", "test")]
+        [InlineData("\t\ntest", "test")]
+        [InlineData("\n\ntest", "test")]
+        [InlineData("test\nline2", "test\nline2")]
+        [InlineData("test\rline2", "test\nline2")]
+        [InlineData("test\r\nline2", "test\nline2")]
+        [InlineData("test\r\r\nline2", "test\n\nline2")]
+        [InlineData("test\r\n\nline2", "test\n\nline2")]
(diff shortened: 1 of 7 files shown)


Shane32:
I merged this as there has not been any comments on it in a week.

src/GraphQLParser.Tests/Validation/LexerValidationTests.cs
```diff
@@ -391,5 +405,33 @@ public void Lex_UnterminatedStringWithText_ThrowsExceptionWithCorrectMessage()
...
             exception.Column.ShouldBe(14);
         }
+
+        [Fact]
+        public void Lex_UnterminatedBlockString_ThrowsExceptionWithCorrectMessage()
+        {
+            var exception = Should.Throw<GraphQLSyntaxErrorException>(() => "\"\"\"".Lex());
+
+            exception.Message.ShouldBe(
+                "Syntax Error GraphQL (1:4) Unterminated string.\n" +
```
sungam3r, line 415: @Shane32  I propose to change error message to `Unterminated block string`. Also you used `BlockString` term in other message. I propose to change it to `block string` since spec does not define term `BlockString`. It defines `BlockStringCharacter` and `BlockStringValue` function.

sungam3r: изображение
the only place where it is mentioned

Shane32: I don't think anyone cares, but sure, go ahead and make a PR if you like.  (I personally think it's fine as-is.)

sungam3r:
Great job! I will make PR with cosmetic edits.

src/GraphQLParser.Tests/Validation/LexerValidationTests.cs
sungam3r: https://github.com/graphql/graphql-spec/issues/899

PR #132, opened 2021-06-28
Add support for description parsing

Shane32:
Adds support for description parsing according to spec.  Requires #131 to be merged in order for tests to pass.

Note in the case where there is a comment both before and after the description, such as in the example below, the second comment is attached to the type definition, and the first comment is added to the list of unattached comments.  If a comment only in one of the two locations, then it is attached to the type definition.  This mechanic is covered by the tests.
```gql
# comment 1
"""
This is my custom scalar
"""
# comment 2
scalar MyScalar
```

Description fields have been added only to the nodes that allow a description.  Since typical GraphQL queries and mutations cannot contain any nodes that would have a description field, and therefor no additional memory could be allocated, there is no need for an additional option to skip parsing of descriptions.

### DIFF
--- a/src/GraphQLParser/AST/GraphQLDescription.cs
+++ b/src/GraphQLParser/AST/GraphQLDescription.cs
@@ -1,0 +1,9 @@
+namespace GraphQLParser.AST
+{
+    public class GraphQLDescription : ASTNode
+    {
+        public override ASTNodeKind Kind => ASTNodeKind.Description;
+
+        public ROM Value { get; set; }
+    }
+}
--- a/src/GraphQLParser/AST/IHasDescription.cs
+++ b/src/GraphQLParser/AST/IHasDescription.cs
@@ -1,0 +1,7 @@
+namespace GraphQLParser.AST
+{
+    public interface IHasDescription
+    {
+        GraphQLDescription? Description { get; set; }
+    }
+}
--- a/src/GraphQLParser.ApiTests/GraphQL-Parser.approved.txt
+++ b/src/GraphQLParser.ApiTests/GraphQL-Parser.approved.txt
@@ -47,6 +47,7 @@ namespace GraphQLParser.AST
         TypeExtensionDefinition = 35,
         DirectiveDefinition = 36,
         Comment = 37,
+        Description = 38,
     }
     public class GraphQLArgument : GraphQLParser.AST.ASTNode, GraphQLParser.AST.INamedNode
     {
@@ -61,6 +62,12 @@ namespace GraphQLParser.AST
         public override GraphQLParser.AST.ASTNodeKind Kind { get; }
         public GraphQLParser.ROM Text { get; set; }
         public override GraphQLParser.AST.GraphQLLocation Location { get; set; }
+    }
+    public class GraphQLDescription : GraphQLParser.AST.ASTNode
+    {
+        public GraphQLDescription() { }
+        public override GraphQLParser.AST.ASTNodeKind Kind { get; }
+        public GraphQLParser.ROM Value { get; set; }
     }
     public class GraphQLDirective : GraphQLParser.AST.ASTNode, GraphQLParser.AST.INamedNode
     {
@@ -280,10 +287,10 @@ namespace GraphQLParser.AST
         protected GraphQLTypeDefinition() { }
         public GraphQLParser.AST.GraphQLName? Name { get; set; }
     }
-    public abstract class GraphQLTypeDefinitionWithDescription : GraphQLParser.AST.GraphQLTypeDefinition
+    public abstract class GraphQLTypeDefinitionWithDescription : GraphQLParser.AST.GraphQLTypeDefinition, GraphQLParser.AST.IHasDescription
     {
         protected GraphQLTypeDefinitionWithDescription() { }
-        public GraphQLParser.AST.GraphQLScalarValue? Description { get; set; }
+        public GraphQLParser.AST.GraphQLDescription? Description { get; set; }
     }
     public class GraphQLTypeExtensionDefinition : GraphQLParser.AST.GraphQLTypeDefinition
     {
@@ -315,6 +322,10 @@ namespace GraphQLParser.AST
         public override GraphQLParser.AST.ASTNodeKind Kind { get; }
         public GraphQLParser.AST.GraphQLType? Type { get; set; }
         public GraphQLParser.AST.GraphQLVariable? Variable { get; set; }
+    }
+    public interface IHasDescription
+    {
+        GraphQLParser.AST.GraphQLDescription? Description { get; set; }
     }
     public interface IHasDirectivesNode
     {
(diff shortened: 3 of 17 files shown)


Shane32:
I have incorporated ideas from #70 into this PR.  If there are no comments I will merge this PR.

sungam3r:
@Shane32 I begin to review this PR but I already suspect what it will affect:
изображение
I believe that this code was to be changed to take new Description syntax node into account, right?

Shane32:
I can't remember exactly, but I think I made changes that would be both non-breaking (i.e. support comments as descriptions) and also support the spec (i.e. support strings/blockstrings as descriptions).  Precedence is to spec - a string/blockstring is a description.  If no string/blockstring, then resort to old non-spec behavior.

Per spec, I believe comments should be completely ignored.  Perhaps we change that in v5.

src/GraphQLParser/AST/GraphQLDescription.cs
```diff
@@ -0,0 +1,9 @@
+namespace GraphQLParser.AST
+{
+    public class GraphQLDescription : ASTNode
```
sungam3r, line 3: @Shane32  Where is `Location` for `GraphQLDescription` syntax node ?

sungam3r:
Yep, I see
изображение

src/GraphQLParser/AST/IHasDescription.cs
```diff
@@ -0,0 +1,7 @@
+namespace GraphQLParser.AST
+{
+    public interface IHasDescription
```
sungam3r, line 3: Well, better to name as `IHasDescriptionNode` to align with `IHasDirectivesNode`.

src/GraphQLParser/AST/GraphQLDescription.cs
Shane32: I think I didn't put one in because the syntax for the parent node technically includes the description, and so the start of the description is the same as the start of the type definition node.  At least that's maybe why I did it???  It's been 4 months so I can't remember my reasoning exactly.

Maybe it should be in there anyway though, so that there's a way to determine the end of the description node?  I guess that would be a new class called `GraphQLDescriptionFull` ... ??

sungam3r: > the syntax for the parent node technically includes the description, and so the start of the description is the same as the start of the type definition node. 

Nope. Those are different syntax nodes with different locations. `GraphQLComment` node has `Location` for example.

> I guess that would be a new class called `GraphQLDescriptionFull`

Yep. Maybe new, maybe existing. I should remember why I did only one `GraphQLComment` class without `GraphQLCommentFull` counterpart. I have already started review, so I can do all the necessary edits myself.

Shane32: image

Shane32: Pretty sure the library is already written this way -- to match the spec.  So a type definition node will have its Location start and end points exactly as described by the spec and shown above.

I don't see any reason why a description node can't have a Location also

src/GraphQLParser/AST/IHasDescription.cs
Shane32: Yes, or vice versa.

src/GraphQLParser/AST/GraphQLDescription.cs
sungam3r: @Shane32 I burst between two options - do all the necessary edits and improvements or only part. In the first case there will be backward incompatible changes: https://github.com/graphql-dotnet/parser/issues/141. Not so serious, but still backward incompatible. In the second case, compatibility will preserved, but the code will be far from perfect. Also I noticed that you missed description for `GraphQLSchemaDefinition`. I want to remove `GraphQLTypeDefinitionWithDescription` class, rename `IHasDescription` and make `IgnoreOptions` as flags enum:
```
None = 0,
Locations = 1,
Comments = 2,
All = Locations | Comments,
```

Personally, I would prefer to release v8 and continue to working on GraphQL v5.

sungam3r: > Pretty sure the library is already written this way -- to match the spec. So a type definition node will have its Location start and end points exactly as described by the spec and shown above.

I have not yet thought about it.

> I don't see any reason why a description node can't have a Location also

Yes, I will add a property.

Shane32: > I want to remove GraphQLTypeDefinitionWithDescription class

Why?

sungam3r: This is not a real class. It only creates an additional level in the inheritance chain. In fact, the property could be added to the basic class `GraphQLTypeDefinition`. Also `GraphQLSchemaDefinition` class **should** have this property but schema definition does not inherit from `GraphQLTypeDefinition`. It would be strange to make one more `WithDescription` base class. Only `GraphQLTypeExtensionDefinition` has no description due to the spec. Nothing terrible is that this property will appear in this class. We do not set it.

Shane32: It allocates extra memory.  Everything we have done in GraphQL.NET and the parser library has all been with a goal of creating no extra allocations.  Why if we are not going to carry through?

If nothing else, it should be set up with a property with no backing field, and then change the 'withdescription' version as an internal class that has a backing field for the property, as is already done for comments.

sungam3r: It allocates only for `GraphQLTypeExtensionDefinition`.

> and then change the 'withdescription' version as an internal class that has a backing field for the property, as is already done for comments.

I thought about it. At first I wanted to go this way. Then I refused it - combinations appear. Too difficult to read/maintain.

No big deal especially since only server parts of the schema may have descriptions and parsing in ordinary scenarios is a one-time operation.

Shane32: There is no added complexity. It is not complex. It is not difficult to write. In fact you simply don’t need to remove optimized functionality that already exists in the library. I do not see any reason to take something optimized and unoptimize it. It does not make sense.

Shane32: The current implementation adds a couple lines of code to the entire library. As in like simply 2-3 lines of code. That is not complex. It is not harder to read. There is no reason to deoptimize the library for the sake of 2 lines of code.

sungam3r: I will think. There is a small chance that schema extensions can also have descriptions (although it is unlikely) - https://github.com/graphql/graphql-spec/issues/900

sungam3r: Well, everything turned out to be simple and difficult at the same time. See http://spec.graphql.org/October2021/#sec-Type-Extensions or even more old http://spec.graphql.org/June2018/#sec-Type-Extensions :
изображение

Now parser supports (partially) only ObjectTypeExtension. Wow! :) No support for other extensions at all, no tests. The good news is that Description property will not exist on GraphQLTypeExtensionDefinition when it stops inherit from `GraphQLTypeDefinition`. All extensions **should** have a separate new base class. Also we should add a bunch of new classes to cover each kind of extension. Looks like a good feature for v8. So I suggest to merge this as is (after your review of course) and continue to work on separate issue - https://github.com/graphql-dotnet/parser/issues/143.

sungam3r: #154

PR #140, opened 2021-11-05
Small changes for string parsing

sungam3r:
This PR represents a result for the review of #131 . Great job @Shane32 . I did not thoroughly check the entire written code. I confess that it is difficult to read. Many counters, many branches. Nevertheless, the general meaning is well understood. Excellent that you wrote all those tests. In addition to cosmetic edits and comments, I reduced the size of the buffers used on the stack. I do not think that we needed such large buffers.

### DIFF
--- a/src/GraphQLParser/LexerContext.cs
+++ b/src/GraphQLParser/LexerContext.cs
@@ -183,7 +183,8 @@ namespace GraphQLParser
         //
         // 1) not too long
         // 2) has no escape sequences
-        // 3) has no '\r` characters
+        // 3) has no '\r' characters
+        // 4) has no initial whitespace on each line, ignoring the first line (or, has no '\n' characters)
         //
         // In this case, ROM for the returned token represents unmodified part of the source ROM,
         // so it can be just sliced from '_source' as you can see in more simple ReadString method.
--- a/src/GraphQLParser/LexerContext.cs
+++ b/src/GraphQLParser/LexerContext.cs
@@ -48,14 +48,9 @@ namespace GraphQLParser
 
             if (code == '"')
             {
-                if (_currentIndex + 2 < _source.Length && _source.Span[_currentIndex + 1] == '"' && _source.Span[_currentIndex + 2] == '"')
-                {
-                    return ReadBlockString();
-                }
-                else
-                {
-                    return ReadString();
-                }
+                return _currentIndex + 2 < _source.Length && _source.Span[_currentIndex + 1] == '"' && _source.Span[_currentIndex + 2] == '"'
+                    ? ReadBlockString()
+                    : ReadString();
             }
 
             return Throw_From_GetToken2(code);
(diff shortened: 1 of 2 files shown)


src/GraphQLParser/LexerContext.cs
```diff
@@ -181,16 +179,24 @@ private Token ReadComment()
             );
         }
 
+        // TODO: this method can still be optimized no not allocate at all if block string:
```
sungam3r, line 182: @Shane32 I am 100% sure that this can be done, but I was not so brave to do it myself :).

src/GraphQLParser/LexerContext.cs
```diff
@@ -181,16 +179,24 @@ private Token ReadComment()
...
+        //
+        // 1) not too long
+        // 2) has no escape sequences
+        // 3) has no '\r` characters
```
Shane32, line 186:
```suggestion
        // 1) not too long
        // 2) has no escape sequences
        // 3) has no '\r' characters
        // 4) has no initial whitespace on each line, ignoring the first line (or, has no '\n' characters)
```

Shane32: Keep in mind that:

`\"""  test\"""` equals `"  test"` but `\"""\n  test\"""` equals `"\ntest"` (whitespace removed from second sample)

Shane32 (changes requested):

Shane32:
> Great job @Shane32 . 

Thanks!  The spec is fairly tricky to implement correctly in an optimized manner.  But I opted for optimization rather than using LINQ or something more readable.  I tried to put some comments in the code to help make it understandable.

Looking back at it now, I don't like that it uses try..catch to catch an overflow and then handle it.  As a general rule of thumb I try not to let any code both generate and catch an exception for two reasons: it significantly slows down execution of code that does not generate an error, and (2) while debugging user code with 'just my code' turned off, the debugger may break on the exception when no error has truly occurred.

If you agree, perhaps you want to modify the PR... up to you.

src/GraphQLParser/LexerContext.cs
sungam3r:
```suggestion
        // 3) has no '\r' characters
```

Shane32 (approved):

sungam3r:
> Looking back at it now, I don't like that it uses try..catch to catch an overflow and then handle it

Your `ReadBlockString` does not use try/catch, `ReadComment/ReadString` do.

Shane32:
> > Looking back at it now, I don't like that it uses try..catch to catch an overflow and then handle it
> 
> Your `ReadBlockString` does not use try/catch, `ReadComment/ReadString` do.

Oh.  Didn't think I would have done that.  Well, sometime we can make an improvement there.

sungam3r:
In fact, I would not worry because of a potential exception handling slowdown. If an exception is happening, then not so often. In addition, the code performed under the `try/catch` block can work faster so without this block. I once read about such compiler issue.

sungam3r:
Waiting a bit for https://github.com/graphql/graphql-spec/issues/899 to verify naming.

Shane32:
It just really annoys me when I go to debug something, and VS breaks on some internal line of code, which isn't an error.  Which just reminds me that there's a significant performance problem whenever that bit of code gets hit.

sungam3r:
I know and understand.

sungam3r:
@Shane32  see https://github.com/graphql/graphql-spec/issues/899#issuecomment-962269529 .  Nevertheless, I will keep `string` and `block string` naming in error messages since there is no `no-block-string` term, only `BlockString` and `StringValue` that **includes** `BlockString`.

PR #142, opened 2021-11-06
Descriptions/Comments/Locations rework

### LINKED ISSUE #141: v8 changes
1) Rename IHasDescription to IHasDescriptionNode
2) IgnoreOptions -> flags enum

sungam3r:
See #132 as a starting point. Fixes #141. Also now parser respects _ignore-locations_ option for comments. Also fixes exceptions caused by placing comments in "unexpected" positions in query/SDL.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Setup .NET Core 5.0 SDK
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: '5.0.201'
+          dotnet-version: '5.0.x'
           source-url: https://nuget.pkg.github.com/graphql-dotnet/index.json
         env:
           NUGET_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
--- a/src/GraphQLParser.Tests/GraphQLAstVisitorTests.cs
+++ b/src/GraphQLParser.Tests/GraphQLAstVisitorTests.cs
@@ -54,8 +54,8 @@ namespace GraphQLParser.Tests
 
         [Theory]
         [InlineData(IgnoreOptions.None)]
-        [InlineData(IgnoreOptions.IgnoreComments)]
-        [InlineData(IgnoreOptions.IgnoreCommentsAndLocations)]
+        [InlineData(IgnoreOptions.Comments)]
+        [InlineData(IgnoreOptions.All)]
         public void Visit_BooleanValueArgument_VisitsOneBooleanValue(IgnoreOptions options)
         {
             using var d = "{ stuff(id : true) }".Parse(new ParserOptions { Ignore = options });
@@ -66,8 +66,8 @@ namespace GraphQLParser.Tests
 
         [Theory]
         [InlineData(IgnoreOptions.None)]
-        [InlineData(IgnoreOptions.IgnoreComments)]
-        [InlineData(IgnoreOptions.IgnoreCommentsAndLocations)]
+        [InlineData(IgnoreOptions.Comments)]
+        [InlineData(IgnoreOptions.All)]
         public void Visit_DefinitionWithSingleFragmentSpread_VisitsFragmentSpreadOneTime(IgnoreOptions options)
         {
             using var d = "{ foo { ...fragment } }".Parse(new ParserOptions { Ignore = options });
@@ -78,8 +78,9 @@ namespace GraphQLParser.Tests
 
         [Theory]
         [InlineData(IgnoreOptions.None)]
-        [InlineData(IgnoreOptions.IgnoreComments)]
-        [InlineData(IgnoreOptions.IgnoreCommentsAndLocations)]
+        [InlineData(IgnoreOptions.Comments)]
+        [InlineData(IgnoreOptions.Locations)]
+        [InlineData(IgnoreOptions.All)]
         public void Visit_DefinitionWithSingleFragmentSpread_VisitsNameOfPropertyAndFragmentSpread(IgnoreOptions options)
         {
             using var d = "{ foo { ...fragment } }".Parse(new ParserOptions { Ignore = options });
@@ -90,8 +91,9 @@ namespace GraphQLParser.Tests
 
         [Theory]
         [InlineData(IgnoreOptions.None)]
-        [InlineData(IgnoreOptions.IgnoreComments)]
-        [InlineData(IgnoreOptions.IgnoreCommentsAndLocations)]
+        [InlineData(IgnoreOptions.Comments)]
+        [InlineData(IgnoreOptions.Locations)]
+        [InlineData(IgnoreOptions.All)]
         public void Visit_DirectiveWithVariable_VisitsVariableOnce(IgnoreOptions options)
         {
             using var d = "{ ... @include(if : $stuff) { field } }".Parse(new ParserOptions { Ignore = options });
@@ -102,8 +104,9 @@ namespace GraphQLParser.Tests
 
         [Theory]
         [InlineData(IgnoreOptions.None)]
-        [InlineData(IgnoreOptions.IgnoreComments)]
-        [InlineData(IgnoreOptions.IgnoreCommentsAndLocations)]
+        [InlineData(IgnoreOptions.Comments)]
+        [InlineData(IgnoreOptions.Locations)]
+        [InlineData(IgnoreOptions.All)]
         public void Visit_EnumValueArgument_VisitsOneEnumValue(IgnoreOptions options)
         {
             using var d = "{ stuff(id : TEST_ENUM) }".Parse(new ParserOptions { Ignore = options });
@@ -114,8 +117,9 @@ namespace GraphQLParser.Tests
 
         [Theory]
         [InlineData(IgnoreOptions.None)]
-        [InlineData(IgnoreOptions.IgnoreComments)]
-        [InlineData(IgnoreOptions.IgnoreCommentsAndLocations)]
+        [InlineData(IgnoreOptions.Comments)]
+        [InlineData(IgnoreOptions.Locations)]
+        [InlineData(IgnoreOptions.All)]
         public void Visit_FloatValueArgument_VisitsOneFloatValue(IgnoreOptions options)
         {
             using var d = "{ stuff(id : 1.2) }".Parse(new ParserOptions { Ignore = options });
@@ -126,8 +130,9 @@ namespace GraphQLParser.Tests
 
         [Theory]
         [InlineData(IgnoreOptions.None)]
-        [InlineData(IgnoreOptions.IgnoreComments)]
-        [InlineData(IgnoreOptions.IgnoreCommentsAndLocations)]
+        [InlineData(IgnoreOptions.Comments)]
+        [InlineData(IgnoreOptions.Locations)]
+        [InlineData(IgnoreOptions.All)]
         public void Visit_FragmentWithTypeCondition_VisitsFragmentDefinitionOnce(IgnoreOptions options)
         {
             using var d = "fragment testFragment on Stuff { field }".Parse(new ParserOptions { Ignore = options });
@@ -138,8 +143,9 @@ namespace GraphQLParser.Tests
 
         [Theory]
         [InlineData(IgnoreOptions.None)]
-        [InlineData(IgnoreOptions.IgnoreComments)]
-        [InlineData(IgnoreOptions.IgnoreCommentsAndLocations)]
+        [InlineData(IgnoreOptions.Comments)]
+        [InlineData(IgnoreOptions.Locations)]
+        [InlineData(IgnoreOptions.All)]
         public void Visit_FragmentWithTypeCondition_VisitsTypeConditionOnce(IgnoreOptions options)
         {
             using var d = "fragment testFragment on Stuff { field }".Parse(new ParserOptions { Ignore = options });
@@ -150,8 +156,9 @@ namespace GraphQLParser.Tests
 
         [Theory]
         [InlineData(IgnoreOptions.None)]
-        [InlineData(IgnoreOptions.IgnoreComments)]
-        [InlineData(IgnoreOptions.IgnoreCommentsAndLocations)]
+        [InlineData(IgnoreOptions.Comments)]
+        [InlineData(IgnoreOptions.Locations)]
+        [InlineData(IgnoreOptions.All)]
         public void Visit_InlineFragmentWithDirectiveAndArgument_VisitsArgumentsOnce(IgnoreOptions options)
         {
             using var d = "{ ... @include(if : $stuff) { field } }".Parse(new ParserOptions { Ignore = options });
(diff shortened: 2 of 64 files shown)


.github/workflows/test.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - name: Setup .NET Core 5.0 SDK
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: '5.0.201'
+          dotnet-version: '5.0.x'
```
sungam3r, line 44: @Shane32 Do you remember why we install .NET Core 3.1 SDK ?

Shane32: Tests run against both 3.1 and 5.0. I wouldn’t change that as 5.0 is not a LTS version.

src/GraphQLParser/ParserContext.Parse.cs
```diff
@@ -250,7 +249,7 @@ private GraphQLDirectiveDefinition ParseDirectiveDefinition()
             ExpectKeyword("on");
             var locations = ParseDirectiveLocations();
 
-            return _ignoreOptions == IgnoreOptions.IgnoreCommentsAndLocations
+            return _ignoreOptions.HasFlag(IgnoreOptions.Locations)
```
Shane32, line 252: All of these conditions are wrong.  If only 'Locations' is specified to be skipped, 'Comments' are also skipped because the non-"full" version is used which does not have a field to store the comments node.

```suggestion
            return _ignoreOptions == IgnoreOptions.All
```

Shane32, line 252: Many of these conditions are wrong.  For instance, here if only 'Locations' is specified to be skipped, 'Comments' are also skipped because the non-"full" version is used which does not have a field to store the comments node.

```suggestion
            return _ignoreOptions == IgnoreOptions.All
```

src/GraphQLParser/AST/GraphQLComment.cs
```diff
@@ -17,17 +16,22 @@ public class GraphQLComment : ASTNode
...
+    internal class GraphQLCommentFull : GraphQLComment
+    {
+        private GraphQLLocation _location;
+
+        public override GraphQLLocation Location
+        {
+            get => _location;
+            set => _location = value;
+        }
+    }
```
Shane32, line 36: Nowhere else do we have a special class for comments without a location.  Optimization is only for ignoring both comments and locations.

src/GraphQLParser/IgnoreOptions.cs
```diff
@@ -1,23 +1,31 @@
...
+
         /// <summary>
         /// Specifies whether to ignore comments when parsing GraphQL document.
         /// </summary>
-        IgnoreComments = 0,
+        Comments = 1,
 
         /// <summary>
-        /// Specifies whether to ignore comments and token locations when parsing GraphQL document.
+        /// Specifies whether to ignore  token locations when parsing GraphQL document.
```
Shane32, line 22:
```suggestion
        /// Specifies whether to ignore token locations when parsing GraphQL document.
```

src/GraphQLParser/AST/GraphQLTypeDefinition.cs
```diff
@@ -1,16 +1,11 @@
 namespace GraphQLParser.AST
 {
-    public abstract class GraphQLTypeDefinition : ASTNode, INamedNode
+    public abstract class GraphQLTypeDefinition : ASTNode, INamedNode, IHasDescriptionNode
     {
         /// <inheritdoc/>
         public GraphQLName? Name { get; set; }
-    }
 
-    public abstract class GraphQLTypeDefinitionWithDescription : GraphQLTypeDefinition, IHasDescription
```
Shane32, line 9: Ok to remove.

src/GraphQLParser/ParserContext.Parse.cs
Shane32: Tests should be added to ensure that each combination of ignored values is properly ignored or retained.

src/GraphQLParser.Tests/ParserTests.cs
```diff
@@ -1082,9 +1144,14 @@ enum Colors {
 ) on QUERY
 ".Parse(new ParserOptions { Ignore = options });
             var defs = document.Definitions;
-            defs.Count.ShouldBe(7);
+            defs.Count.ShouldBe(8);
             var parseComments = options == IgnoreOptions.None;
```
Shane32, line 1148:
```suggestion
            var parseComments = options == IgnoreOptions.None || options == IgnoreOptions.Locations;
```

Shane32: You may want to check to see a similar change is needed anywhere else in the tests.

src/GraphQLParser.Tests/ParserTests.cs
```diff
@@ -910,9 +960,14 @@ enum Colors {
...
```
Shane32, line 964:
```suggestion
            var parseComments = !options.HasFlag(IgnoreOptions.Comments);
```

src/GraphQLParser.Tests/ParserTests.cs
Shane32: Or this

```suggestion
            var parseComments = !options.HasFlag(IgnoreOptions.Comments);
```

src/GraphQLParser.Tests/GraphQLAstVisitorTests.cs
```diff
@@ -54,8 +54,8 @@ public GraphQLAstVisitorTests()
 
         [Theory]
         [InlineData(IgnoreOptions.None)]
-        [InlineData(IgnoreOptions.IgnoreComments)]
-        [InlineData(IgnoreOptions.IgnoreCommentsAndLocations)]
+        [InlineData(IgnoreOptions.Comments)]
+        [InlineData(IgnoreOptions.All)]
```
Shane32, line 58:
```suggestion
        [InlineData(IgnoreOptions.Comments)]
        [InlineData(IgnoreOptions.Locations)]
        [InlineData(IgnoreOptions.All)]
```
For the sake of consistency, shouldn't Locations be here also or am I missing something?  Also line 69

src/GraphQLParser.Tests/ParserTests.cs
```diff
@@ -47,8 +47,9 @@ public void Extra_Comments_Should_Read_Correctly(IgnoreOptions options)
 
         [Theory]
         //[InlineData(IgnoreOptions.None)]
-        [InlineData(IgnoreOptions.IgnoreComments)]
-        [InlineData(IgnoreOptions.IgnoreCommentsAndLocations)]
+        [InlineData(IgnoreOptions.Comments)]
```
Shane32, line 50: Changes should be made on various tests (e.g. the `Comments_on_FragmentSpread_Should_Read_Correclty` test) to include `[InlineData(IgnoreOptions.Locations)]` to test reading comments when only locations are ignored

Shane32 (changes requested):

src/GraphQLParser/ParserContext.Parse.cs
Shane32: I think actually the problem isn't that the tests don't exist, but that no `[InlineData(IgnoreOptions.Locations)]` were added to tests for comments

src/GraphQLParser.Tests/ParserTests.cs
Shane32: (Not this test, but I couldn't comment on line 69)

Shane32:
Note that GraphQLLocation is a struct, and takes almost no processing power to compute (since the location is already known).  So if there are no optimized classes for classes, then running the parser with only `IgnoreOptions.Locations` is pointless, as it will use the same memory, and probably an identical amount of processing time.  The reason the option existed previously is because when ignoring both comments and locations (neither of which are necessary for execution of a query by GraphQL.NET), optimized classes can be used that save memory requirements of both fields, and the heap allocations and processing power to process comment fields.

I also think (if I remember right) that if an error occurs during parsing, an exception is generated within the parser that utilizes the location of the error, even if location is specified to be ignored when parsing.

So while flags are a better design, it may be a bit misleading to users in that ignoring locations (but not comments) will actually do anything beneficial.  Just because of that reason, I'd probably keep the options values like they were, especially as we have no intention of adding additional flags or adding additional classes to options for ignoring only locations.  Assume that we proceed with switching it to flags, I might add a comment to this effect - either in the xml comment for the enum option, or in the readme file.

src/GraphQLParser/ParserContext.Parse.cs
sungam3r: Well, my mistake is obvious 🤦 .

src/GraphQLParser/AST/GraphQLComment.cs
sungam3r: I renamed `GraphQLCommentFull` into `GraphQLCommentWithLocation`. I will soon make a commit, where I added a lot of new internal classes to align the behavior. All optimizations will be in place. I will need some time to finish and write tests. I want to show in advance what it turns out.

src/GraphQLParser/IgnoreOptions.cs
sungam3r: 👍

src/GraphQLParser.Tests/GraphQLAstVisitorTests.cs
sungam3r: Yes.

src/GraphQLParser.Tests/ParserTests.cs
sungam3r: Yes

sungam3r:
>  So if there are no optimized classes for skipping only locations, then running the parser with only IgnoreOptions.Locations is pointless, as it will use the same memory, and probably an identical amount of processing time
> So while flags are a better design, it may be a bit misleading to users in that ignoring locations (but not comments) will actually do anything beneficial.

I did so that every option gave effect. I have to write tests and put some code for reading comments in 3-4 places. When I worked on this PR, I realized that the processing of comments is a very unpleasant process. They may be everywhere! We are far from properly handling comment. I found that if you put a comment after the name of the arg but before its value, the parser generates an exception. Boom! See `CommentsOnValues.graphql` file. I have to add `ParseComment()` call into more places.

src/GraphQLParser/NodeHelper.cs
```diff
@@ -0,0 +1,416 @@
+using System.Runtime.CompilerServices;
+using GraphQLParser.AST;
+
+namespace GraphQLParser
+{
+    internal static class NodeHelper
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static GraphQLComment CreateGraphQLComment(IgnoreOptions options)
```
sungam3r, line 9: First three methods differ from others. `GraphQLComment`, `GraphQLDocument` and `GraphQLDescription` have no comments. All other methods are completely the same. Perhaps when I will write tests, I will see that comments are not applicable to some other AST nodes.

Shane32:
Why not just discard all comments (according to spec) for v8?  Eliminate the options for parsing it, eliminate the complexity of the classes to hold comments, etc.

> Comments are Ignored like white space and may appear after any token, or before a LineTerminator, and have no significance to the semantic meaning of a GraphQL Document.

Shane32:
> I renamed GraphQLCommentFull into GraphQLCommentWithLocation. I will soon make a commit, where I added a lot of new internal classes to align the behavior. All optimizations will be in place. I will need some time to finish and write tests. I want to show in advance what it turns out.

I really think you are wasting a lot of time here, and making the project unnecessarily complex with "a lot of new internal classes" that are not necessary when the only project that uses this parser -- GraphQL.NET -- only does two things: (a) skip locations & comments for execution, and (b) load comments for backwards compatibility when creating a schema-first schema.  And as for the second point, descriptions have long been in the spec now to be done by strings, not comments.  So even that is not necessary.  And even if 'comments without locations' could be used by the schema-first initialization, memory requirements does not matter because schema initialization is only done once.  (Or should be.)

Shane32:
(Edit: oops GraphQL always uses locations)

We should really eliminate parsing without locations, since GraphQL requires the use of locations in error messages.

sungam3r:
> Why not just discard all comments (according to spec) for v8? Eliminate the options for parsing it, eliminate the complexity of the classes to hold comments, etc.

Of course, we **can** do it. But from the point of view of the syntactic parsing it will be a step back. People may need comments on completely different reasons. The most obvious of them is the documentation provision through the introspection. We will force everyone to use descriptions. At best, consumers will have to change their SDLs. In the worst case, they will not be able to do this at all, because they do not own these schemas. Then they have to change these schemas (SDL) dynamically, which will require  ... the syntactic analysis that we have deleted 😄 .

sungam3r:
> We should really eliminate optimization of parsing without locations, since GraphQL requires the use of locations in error messages.

I'm not sure that this requirement somehow limits us. The specification describes what should occur between the client and the server. Parser is a general mechanism that can work without the presence of any client. For example, it may be the transformation from one SDL to another SDL or to some different file format. Getting SDL statistics, whatever.

Shane32:
None of those scenarios require memory optimization.  Only execution does, which requires location information.

sungam3r:
> when the only project that uses this parser -- GraphQL.NET 

Why so 😄 ? It is better to say "the only **public** project that uses this parser" although I doubt even that it is indeed the only public. I use the parser in my private projects for example. In any case, we need to eliminate exceptions caused by the placing comments in "unexpected" positions.

Shane32:
You can write all that extra code if you want; I wouldn't 🤷‍♂️.

Shane32:
> In any case, we need to eliminate exceptions caused by the placing comments in "unexpected" positions.

For sure.

Shane32:
> I use the parser in my private projects for example.

Where (a) you skip parsing location information, and (b) in situations that require hyper-optimized memory allocations?

Shane32:
Hey, go ahead 👍 ... maybe it's useful to someone; maybe you.

sungam3r:
OK. By the way, if we talk about the complexity of the code, then after the changes made code became easier. I deleted all the ternary operators around AST creation and initializing. All properties are set once now. The number of codes in the "main" parser file decreased. I look positively on changes.

Shane32:
Is it ready for another review?

sungam3r:
Not yet. I'll let you know.

sungam3r:
@Shane32 I'm ready for review. It turned out to be a big one PR.

What you should pay attention to:
1. Single `ParseComment()` call inside `Advance` so now we handle all comments anywhere.
2. All nodes parse and set their comments now.
3. New tests verify additional comments in new places. I can add additional tests for comments in awkward positions like before `:` or `=` tokens. Now all of them should reside into unattached comments. No more exceptions!
4. `ParserContext` ctor calls `CreateGraphQLDocument` and `Advance`.
5. I renamed some public classes to align their names to ones from the spec: GraphQLFieldSelection -> GraphQLField, GraphQLOperationTypeDefinition  -> GraphQLRootOperationTypeDefinition

I also walked through all the `ParseXXX` methods and added comment to each with reference to the corresponding part of the specification (grammar).

sungam3r:
Now it will be easier to work on the new features from the October spec.

Shane32 (approved):

PR #149, opened 2021-11-13
AstVisitor rework

### LINKED ISSUE #11: Is there a way to turn the AST back into a graphql query string?
I'm interested into urning Linq Expressions into graphql queries (e.g. so a .NET client can treat a graphQL endpoint as an IQueryable), and was thinking I could use your AST to do this, but I can't find a way to turn it back into a graphql query string. Is there a way to do this?

sungam3r:
fixes #11 , supersedes #20 

Visitors API was completely rewritten, optimized, tested.


Shane32:
Is this all synchronous code? Synchronous code that writes streams should be deprecated and replaced with async code, just like what was done with System.Text.Json. Certainly we should not be writing new synchronous code here in 2021. And even more certainly if we are replacing the visitor api, it should be a new async api. 

This is a good candidate for using ValueTask, however, to eliminate allocations when writing to a string builder or other non-async stream. Or for any other non-async needs of the visitor api.

sungam3r:
`TextWriter` has no write methods that return `ValueTask`.

sungam3r:
Nevermind, I can change all signatures to `ValueTask`.

src/Directory.Build.props
```diff
@@ -19,7 +19,7 @@
     <!-- https://github.com/clairernovotny/DeterministicBuilds -->
     <ContinuousIntegrationBuild Condition="'$(GITHUB_ACTIONS)' == 'true'">True</ContinuousIntegrationBuild>
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
-    <NoWarn>$(NoWarn);1591</NoWarn> <!--TODO: temporary solution, remove-->
+    <NoWarn>$(NoWarn);1591;IDE1006</NoWarn> <!--TODO: temporary solution, remove-->
```
sungam3r, line 22: `Async` suffix rule

src/GraphQLParser.ApiTests/GraphQL-Parser.approved.txt
```diff
@@ -142,7 +143,7 @@ namespace GraphQLParser.AST
         public System.Collections.Generic.List<GraphQLParser.AST.GraphQLDirective>? Directives { get; set; }
         public override GraphQLParser.AST.ASTNodeKind Kind { get; }
         public GraphQLParser.AST.GraphQLSelectionSet? SelectionSet { get; set; }
-        public GraphQLParser.AST.GraphQLNamedType? TypeCondition { get; set; }
```
sungam3r, line 145: see http://spec.graphql.org/October2021/#sec-Inline-Fragments and http://spec.graphql.org/October2021/#FragmentDefinition  - [TypeCondition](http://spec.graphql.org/October2021/#TypeCondition) node should be added

src/GraphQLParser/AST/GraphQLVariableDefinition.cs
```diff
@@ -1,15 +1,20 @@
+using System.Collections.Generic;
+
 namespace GraphQLParser.AST
 {
-    public class GraphQLVariableDefinition : ASTNode
+    public class GraphQLVariableDefinition : ASTNode, IHasDirectivesNode
     {
-        public object? DefaultValue { get; set; }
```
sungam3r, line 5: @Shane32 I somehow missed this `object` type before. Changed to `GraphQLValue` per the [spec](http://spec.graphql.org/October2021/#VariableDefinition).

src/GraphQLParser/GraphQLParser.csproj
```diff
@@ -11,6 +11,7 @@
 
   <ItemGroup>
     <PackageReference Include="System.Memory" Version="4.5.4" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
```
sungam3r, line 14: For `ValueTask`.

src/GraphQLParser/ParserContext.Parse.cs
```diff
@@ -1149,6 +1144,7 @@ private GraphQLVariableDefinition ParseVariableDefinition()
 
             def.Type = ParseType();
             def.DefaultValue = Skip(TokenKind.EQUALS) ? ParseValueLiteral(true) : null;
+            def.Directives = ParseDirectives();
```
sungam3r, line 1147: [_VariableDefinition_](http://spec.graphql.org/October2021/#VariableDefinition) **may** have directives

src/GraphQLParser/Visitors/SDLWriter.cs
```diff
@@ -0,0 +1,765 @@
...
+            await Visit(nullValue.Comment, context);
+            await context.Writer.WriteAsync("null");
+        }
+
+        /// <inheritdoc/>
+        public override async ValueTask VisitStringValue(GraphQLScalarValue stringValue, TContext context)
+        {
+            await Visit(stringValue.Comment, context);
+            await context.Writer.WriteAsync('\"');
+            await Write(context, stringValue.Value);
```
Shane32, line 584: Needs escaping of various characters; please add test.

Optional: detect if the string is multi-line, and can be formatted as a blockstring without loss of data, and if both are true, print it that way.

src/GraphQLParser/Visitors/SDLWriter.cs
```diff
@@ -0,0 +1,765 @@
...
+                    case '\r':
+                        break;
+
+                    case '\n':
+                        await context.Writer.WriteLineAsync();
+                        needStartNewLine = true;
+                        break;
+
+                    default:
+                        await context.Writer.WriteAsync(current);
```
Shane32, line 98: The sequence `"""` must be escaped.  Please add test.

src/GraphQLParser/Visitors/SDLWriter.cs
```diff
@@ -0,0 +1,765 @@
...
+                        break;
+                }
+            }
+
+            await context.Writer.WriteLineAsync();
+        }
+
+        /// <inheritdoc/>
+        public override async ValueTask VisitDescription(GraphQLDescription description, TContext context)
```
Shane32, line 68: Not all strings can be written in the blockstring format.  For example, a string that starts with a newline.  This method should ideally check to be sure the input string can be formatted as a blockstring without loss of information, and if not, be formatted as a string.

src/GraphQLParser/Visitors/IWriteContext.cs
```diff
@@ -0,0 +1,12 @@
+using System.Collections.Generic;
+using System.IO;
+
+namespace GraphQLParser.Visitors
+{
+    public interface IWriteContext : IVisitorContext
+    {
+        TextWriter Writer { get; }
+
+        Stack<AST.ASTNode> Parent { get; }
```
Shane32, line 10:
```suggestion
        Stack<AST.ASTNode> Parents { get; }
```

src/GraphQLParser/Visitors/IWriteContext.cs
```diff
@@ -0,0 +1,12 @@
...
+    }
```
Shane32, line 11:
```suggestion

        CancellationToken CancellationToken { get; }
    }
```
I suggest adding cancellation token support

src/GraphQLParser/Visitors/SDLWriter.cs
```diff
@@ -0,0 +1,765 @@
...
+            return level;
+        }
+
+        private ValueTask Write(TContext context, ROM value)
+        {
+            var task =
+#if NETSTANDARD2_0
+            context.Writer.WriteAsync(value.ToString());
+#elif NETSTANDARD2_1_OR_GREATER
+            context.Writer.WriteAsync(value);
```
Shane32, line 749:
```suggestion
            context.Writer.WriteAsync(value, context.CancellationToken);
```

Shane32: Cancellation tokens are supported here

src/GraphQLParser/Visitors/StructureWriter.cs
```diff
@@ -0,0 +1,29 @@
...
+        public override async ValueTask Visit(ASTNode? node, TContext context)
+        {
+            if (node == null)
+                return;
+
+            for (int i = 0; i < context.Parent.Count; ++i)
+                await context.Writer.WriteAsync("  ");
+
+            context.Parent.Push(node);
+            await context.Writer.WriteLineAsync(node.Kind.ToString());
```
Shane32, line 24: If we can convert a string to a ROM, we can also use a cancellation token here for .NET Standard 2.1.

src/GraphQLParser/Visitors/SDLWriter.cs
```diff
@@ -0,0 +1,765 @@
...
+                {
+                    await WriteIndent(context, level);
+                    needStartNewLine = false;
+                }
+
+                char current = description.Value.Span[i];
+                switch (current)
+                {
+                    case '\r':
+                        break;
```
Shane32, line 90: Note: with this logic, description strings encoded with CRLF will result in strings being encoded as LF after being decoded.  I suggest adding a comment within the method that this change may take place -- or better yet, encode strings containing CRs as strings instead of blockstrings.  Tests should be added either way.

src/GraphQLParser/Visitors/SDLWriter.cs
```diff
@@ -0,0 +1,765 @@
...
+
+                    case '\n':
+                        await context.Writer.WriteLineAsync();
+                        needStartNewLine = true;
+                        break;
+
+                    default:
+                        await context.Writer.WriteAsync(current);
```
Shane32, line 59: It would be nice to use the cancellationtoken within this method... but it's an optimization that can be done later.

src/GraphQLParser/Visitors/SDLWriter.cs
Shane32: Might want to add a `private ValueTask WriteLine(TContext context)` method so the writeline methods can utilize the cancellation token.  (Store `Environment.NewLine` as a ROM statically, etc)

Shane32: And maybe a `private ValueTask WriteLine(char char)` method.....

Shane32 (changes requested):
This is looking great!  Here's a few comments, mostly about escaping strings and adding some cancellation token support.

Shane32:
By the way, should we be adding `.ConfigureAwait(false)` everywhere?

https://devblogs.microsoft.com/dotnet/configureawait-faq/

src/GraphQLParser/Visitors/IWriteContext.cs
sungam3r: At first I called the property in this way.

sungam3r:
> By the way, should we be adding .ConfigureAwait(false) everywhere?

Of course.

src/GraphQLParser/Visitors/SDLWriter.cs
sungam3r: > and can be formatted as a blockstring without loss of data

Example ?

src/GraphQLParser/Visitors/SDLWriter.cs
sungam3r: > and if not, be formatted as a string.

One-line string? Or are you talking about change description node into comment ?

src/GraphQLParser/Visitors/IWriteContext.cs
sungam3r: done

src/GraphQLParser/Visitors/IWriteContext.cs
sungam3r: done

src/GraphQLParser/Visitors/SDLWriter.cs
sungam3r: 😄 I read these comments after I fixed all that stuff. Check again `WriteContextExtensions`.
Regarding writing single char. I rejected a separate method and pass a string from one symbol. Not such a big trouble, but the code is slightly clearer and unified.

src/GraphQLParser/Visitors/StructureWriter.cs
sungam3r: 😄 Already did.

src/GraphQLParser/Visitors/SDLWriter.cs
sungam3r: > strings instead of blockstrings

I'm still confused about string vs blockstring, can you give an example?

By the way I want to rewrite both methods for comment and description node. Now I added the `Slice` call to get around the problem:
```
CS4012	Parameters or locals of type 'ReadOnlySpan<char>' cannot be declared in async methods or async lambda expressions.
```

In addition, I do not really like this char-by-char transmission. At first I wrote a method that finds the chain of non-`\r\n` chars and writes those spans into writer, but it was rather complicated so I swiched to char-by-char.

src/GraphQLParser/Visitors/SDLWriter.cs
sungam3r: Done, all writes into `TextWriter` now are handled via extension methods respecting token from context.

sungam3r:
@Shane32 I have left to solve the issues with strings/blockstrings encoding. I'll do it tomorrow or in the coming days.

sungam3r:
What is done for now:

1. General design for visitors, naming
2. Full-async (for netstandard2.1 100%)
3. Cancellation support
4. Some bugs were fixed, some new nodes/properties were added

src/GraphQLParser/Visitors/SDLWriter.cs
sungam3r: I took test data from lexer tests. This is your test with 30 cases. I changed some values. Honestly, work with escape symbols is very tired. I would like to finish with this PR and go on. I could be mistaken somewhere, further use will show. We will add tests. Or you can add additional test cases yourself where you think right.

src/GraphQLParser/Visitors/SDLWriter.cs
sungam3r: done

src/GraphQLParser/Visitors/SDLWriter.cs
sungam3r: See https://github.com/graphql-dotnet/parser/pull/149#discussion_r748885168

Please show an example here or add a test case yourself. I'm a little confused with all this escaping stuff.

sungam3r:
@Shane32 Waiting for examples.

Shane32 (changes requested):
The only major issue I see is that string values are not escaped.  You already wrote code within VisitDescription to encode strings; just copy or reuse the code block and it should be fine.  I can do a detailed review of the encoding for strings and blockstrings at a later time and add another PR if I see any other suggestions.  Ok?

src/Directory.Build.props
Shane32: Just wondering why you decided to have the Visit methods not end in Async?

src/GraphQLParser/Visitors/SDLWriter.cs
Shane32: I agree to finish the PR and go on.  The API will not change if we improve these methods in a later PR, either to fix an issue or enhance performance.

src/GraphQLParser/Visitors/SDLWriter.cs
```diff
@@ -0,0 +1,847 @@
...
+            await Visit(nullValue.Comment, context).ConfigureAwait(false);
+            await context.Write("null").ConfigureAwait(false);
+        }
+
+        /// <inheritdoc/>
+        public override async ValueTask VisitStringValue(GraphQLScalarValue stringValue, TContext context)
+        {
+            await Visit(stringValue.Comment, context).ConfigureAwait(false);
+            await context.Write("\"").ConfigureAwait(false);
+            await context.Write(stringValue.Value).ConfigureAwait(false);
```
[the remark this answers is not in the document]
sungam3r: Thanks, understood. I did so and added one new test.

src/Directory.Build.props
sungam3r: Year or two ago I would do so, yes. I know about all those recommendations and styles. I control several hundred repos at work unifiing code style/design and this have always been one of the permanent questions. My intuition suggests me that if the API is initially designed as (completely) asynchronous, then it makes no sense to append `Async` everywhere. It only creates additional noise. By the way, I remove that suppression from dir.props, see .editorconfig file.

src/GraphQLParser/Visitors/SDLWriter.cs
sungam3r: I'm OK to wait 1-2-3 days for your review. At least until I start a new branch working on type extensions. I am 80% sure that I have errors in formatting/edge cases.

sungam3r:
ping @Shane32 , any suggestions/comments/test cases?

Shane32:
Not at this time.

Shane32 (approved):

PR #152, opened 2021-12-29
Auto approve API

sungam3r:


src/GraphQLParser/GraphQLParser.csproj
```diff
@@ -5,7 +5,6 @@
     <Product>GraphQL Parser</Product>
     <AssemblyTitle>GraphQL Parser for .NET</AssemblyTitle>
     <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
-    <AssemblyName>GraphQL-Parser</AssemblyName>
```
sungam3r, line 8: It caused problems in API tests.

Shane32 (approved):

src/GraphQLParser/GraphQLParser.csproj
sungam3r: Fixed in #156. Should be <PackageId>

PR #153, opened 2022-01-02
Add maximum depth limit to Parser

Shane32:
An unlimited depth can result in a stack overflow exception when parsing long queries.

### DIFF
--- a/src/GraphQLParser.Tests/ParserTests.cs
+++ b/src/GraphQLParser.Tests/ParserTests.cs
@@ -15,7 +15,7 @@ namespace GraphQLParser.Tests
         [Fact]
         public void Should_Throw_With_Deep_Query()
         {
-            var count = 24000;
+            var count = 200;
             var sb = new System.Text.StringBuilder(count * 3);
             for (int i = 0; i < count; i++)
                 sb.Append("{a");
@@ -27,7 +27,7 @@ namespace GraphQLParser.Tests
         [Fact]
         public void Should_Throw_With_Deep_Literal()
         {
-            var count = 24000;
+            var count = 200;
             var sb = new System.Text.StringBuilder(count * 4 + 10);
             sb.Append("{a(b:");
             for (int i = 0; i < count; i++)
--- a/src/GraphQLParser.Tests/ParserTests.cs
+++ b/src/GraphQLParser.Tests/ParserTests.cs
@@ -67,6 +67,19 @@ namespace GraphQLParser.Tests
         }
 
         [Fact]
+        public void Should_Parse_With_Shallow_Long_Query()
+        {
+            var count = 200;
+            var sb = new System.Text.StringBuilder(count * 5);
+            sb.Append('{');
+            for (int i = 0; i < count; i++)
+                sb.Append(" a" + i);
+            sb.Append('}');
+            var query = sb.ToString();
+            _ = query.Parse();
+        }
+
+        [Fact]
         public void Should_Throw_With_MaxDepth_0_On_SimpleQuery()
         {
             var query = "{a}";
--- a/src/GraphQLParser.Tests/ParserTests.cs
+++ b/src/GraphQLParser.Tests/ParserTests.cs
@@ -12,6 +12,33 @@ namespace GraphQLParser.Tests
     {
         private static readonly string _nl = Environment.NewLine;
 
+        [Fact]
+        public void Should_Throw_With_Deep_Query()
+        {
+            var count = 24000;
+            var sb = new System.Text.StringBuilder(count * 3);
+            for (int i = 0; i < count; i++)
+                sb.Append("{a");
+            sb.Append(new string('}', count));
+            var query = sb.ToString();
+            Should.Throw<GraphQLSyntaxErrorException>(() => query.Parse()).Description.ShouldBe("Maximum depth exceeded.");
+        }
+
+        [Fact]
+        public void Should_Throw_With_Deep_Literal()
+        {
+            var count = 24000;
+            var sb = new System.Text.StringBuilder(count * 4 + 10);
+            sb.Append("{a(b:");
+            for (int i = 0; i < count; i++)
+                sb.Append("{c:");
+            sb.Append("{}");
+            sb.Append(new string('}', count));
+            sb.Append(")}");
+            var query = sb.ToString();
+            Should.Throw<GraphQLSyntaxErrorException>(() => query.Parse()).Description.ShouldBe("Maximum depth exceeded.");
+        }
+
         [Theory]
         [InlineData(IgnoreOptions.None)]
         //[InlineData(IgnoreOptions.Comments)]
--- a/src/GraphQLParser.Tests/ParserTests.cs
+++ b/src/GraphQLParser.Tests/ParserTests.cs
@@ -21,7 +21,7 @@ namespace GraphQLParser.Tests
                 sb.Append("{a");
             sb.Append(new string('}', count));
             var query = sb.ToString();
-            Should.Throw<GraphQLSyntaxErrorException>(() => query.Parse()).Description.ShouldBe("Maximum depth exceeded.");
+            Should.Throw<GraphQLMaxDepthExceededException>(() => query.Parse());
         }
 
         [Fact]
@@ -36,28 +36,28 @@ namespace GraphQLParser.Tests
             sb.Append(new string('}', count));
             sb.Append(")}");
             var query = sb.ToString();
-            Should.Throw<GraphQLSyntaxErrorException>(() => query.Parse()).Description.ShouldBe("Maximum depth exceeded.");
+            Should.Throw<GraphQLMaxDepthExceededException>(() => query.Parse());
         }
 
         [Fact]
         public void Should_Throw_With_MaxDepth_0_On_SimpleQuery()
         {
             var query = "{a}";
-            Should.Throw<GraphQLSyntaxErrorException>(() => Parser.Parse(query, new ParserOptions { MaxDepth = 0 })).Description.ShouldBe("Maximum depth exceeded.");
+            Should.Throw<GraphQLMaxDepthExceededException>(() => Parser.Parse(query, new ParserOptions { MaxDepth = 0 }));
         }
 
         [Fact]
         public void Should_Throw_With_MaxDepth_0_On_TypeDefinition()
         {
             var query = "scalar Test";
-            Should.Throw<GraphQLSyntaxErrorException>(() => Parser.Parse(query, new ParserOptions { MaxDepth = 0 })).Description.ShouldBe("Maximum depth exceeded.");
+            Should.Throw<GraphQLMaxDepthExceededException>(() => Parser.Parse(query, new ParserOptions { MaxDepth = 0 }));
         }
 
         [Fact]
         public void Should_Throw_With_MaxDepth_1_On_SimpleQuery()
         {
             var query = "{a}";
-            Should.Throw<GraphQLSyntaxErrorException>(() => Parser.Parse(query, new ParserOptions { MaxDepth = 1 })).Description.ShouldBe("Maximum depth exceeded.");
+            Should.Throw<GraphQLMaxDepthExceededException>(() => Parser.Parse(query, new ParserOptions { MaxDepth = 1 }));
         }
 
         [Fact]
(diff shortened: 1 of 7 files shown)


src/GraphQLParser/ParserContext.Parse.cs
```diff
@@ -145,6 +145,7 @@ private ASTNode Throw_From_ParseDefinition()
         private List<ASTNode> ParseDefinitionsIfNotEOF()
         {
             var result = new List<ASTNode>();
+            IncreaseDepth();
```
Shane32, line 148: Probably not needed here, but wasn't sure.

src/GraphQLParser/ParserOptions.cs
```diff
@@ -10,5 +10,11 @@ public struct ParserOptions
         /// By default, all comments are ignored.
         /// </summary>
         public IgnoreOptions Ignore { get; set; }
+
+        /// <summary>
+        /// Maximum recursion depth during parsing.
+        /// Defaults to 64 if not set.
+        /// </summary>
+        public int? MaxDepth { get; set; }
```
Shane32, line 18: Here I have a nullable `int`.  For System.Text.Json, it's just an `int`, and the docs say that a value of 0 defaults to 64.  Should we do that instead?

src/GraphQLParser/ParserOptions.cs
```diff
@@ -10,5 +10,11 @@ public struct ParserOptions
         /// By default, all comments are ignored.
         /// </summary>
         public IgnoreOptions Ignore { get; set; }
+
+        /// <summary>
+        /// Maximum recursion depth during parsing.
```
sungam3r, line 15:
```suggestion
        /// Maximum allowed recursion depth during parsing.
```

src/GraphQLParser/ParserOptions.cs
sungam3r: From here https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonreaderoptions.maxdepth?view=netcore-3.1 ?

sungam3r:
Interesting PR :)

src/GraphQLParser/ParserContext.cs
```diff
@@ -37,6 +41,23 @@ public ParserContext(ROM source, ParserOptions options)
...
+        private void IncreaseDepth()
+        {
+            // Encourage compiler inlining of this method by moving exception to a separate method
+            if (_currentDepth++ >= _maxDepth)
+                ThrowMaxDepthException();
+        }
+
+        private void ThrowMaxDepthException()
+        {
+            throw new GraphQLSyntaxErrorException("Maximum depth exceeded.", _source, _currentToken.Start);
```
sungam3r, line 53: It's not a syntax exception though we _can_ use `_source` and `_currentToken`. `InvalidOperationException` suits better for me.

src/GraphQLParser/ParserOptions.cs
sungam3r: I would use nullable.

src/GraphQLParser/ParserContext.cs
```diff
@@ -37,6 +41,23 @@ public ParserContext(ROM source, ParserOptions options)
             Advance();
         }
 
+        private void IncreaseDepth()
+        {
+            // Encourage compiler inlining of this method by moving exception to a separate method
+            if (_currentDepth++ >= _maxDepth)
```
sungam3r, line 47:
```suggestion
            if (++_currentDepth >= _maxDepth)

```

Please add test with max depth = 0. It should work on simple values.

src/GraphQLParser/ParserContext.cs
Shane32: I can create a new exception for it that inherits from `GraphQLSyntaxErrorException`.  Or we can create a new parent exception to cover both child exceptions.  But it should be a parser exception just as exceeding the maximum depth during json parsing throws a `JsonException`.

Also, I would like to maintain binary compatibility within this PR.  Then it can pulled in via nuget by users immediately, fixing the vulnerability within GraphQL.NET.

src/GraphQLParser/ParserContext.cs
Shane32: Sure I can add a test.  But why the change?

src/GraphQLParser/ParserContext.cs
Shane32: I think we should use `_source` and `_currentToken`.  No reason not to do so.

sungam3r: I'm fine with `GraphQLParserException` base class and `GraphQLSyntaxErrorException: GraphQLParserException` + `GraphQLMaxDepthExceededException: GraphQLParserException`.

src/GraphQLParser/ParserContext.cs
sungam3r: Your test should work on literal values (depth 0) and should not work on something else (depth 1).
`a++` returns a, `++a` returns incremented a, so without suggested fix test will fail I think.

Shane32: As it is now, max depth must be 1 to work on something simple like `scalar Test`.  Max depth must be 2 to work on a basic query like `{a}`.  Seems to make sense to me.  Making the change you suggested would require MaxDepth settings of 2 and 3 respectively.  We can initialize `_currentDepth` to `-1` if we want to, essentially, add one level to the maximum depth setting.

Shane32: > Making the change you suggested would require MaxDepth settings of 2 and 3 respectively.

Tested and confirmed

Shane32: > We can initialize `_currentDepth` to `-1` if we want to, essentially, add one level to the maximum depth setting.

Please review and lmk if I should make this change, so MaxDepth = 0 will parse simple things like `scalar Test`

sungam3r: It depends on what numbering we have - from 0 or from 1.

sungam3r: Worth to write about it in the comments. For me there is no big difference 0 or 1 to choose as "initial ground level". It should simply be noted clearly.

Shane32: Added note "Minimum value is 1` in the xml comment.

src/GraphQLParser/ParserContext.cs
Shane32: Updated as suggested.

sungam3r: Forgot to push?

Shane32: Yup

src/GraphQLParser/Exceptions/GraphQLParserException.cs
```diff
@@ -0,0 +1,120 @@
+using System;
+using System.Linq;
+using System.Text;
+
+namespace GraphQLParser.Exceptions
+{
+    /// <summary>
+    /// An exception representing a GraphQL document syntax error.
```
sungam3r, line 8:
```suggestion
    /// An exception representing a GraphQL document parsing error.
```

src/GraphQLParser/Exceptions/GraphQLMaxDepthExceededException.cs
```diff
@@ -0,0 +1,16 @@
+namespace GraphQLParser.Exceptions
+{
+    /// <summary>
+    /// An exception representing a GraphQL document syntax error.
```
sungam3r, line 4: Copy-pasted comment :(

Shane32:
```suggestion
    /// An exception representing a GraphQL 'maximum depth exceeded' error.
```

Shane32: > Copy-pasted comment :(

Yeah I've been working on way too many PRs today!  Getting lost in the changes...not reviewing my own work...etc

sungam3r: It may seem that this is a standard error from the spec.
```suggestion
    /// An exception representing a 'maximum depth exceeded' parser error.
```

src/GraphQLParser.Tests/ParserTests.cs
```diff
@@ -12,6 +12,68 @@ public class ParserTests
...
+
+        [Fact]
+        public void Should_Throw_With_MaxDepth_0_On_TypeDefinition()
+        {
+            var query = "scalar Test";
+            Should.Throw<GraphQLMaxDepthExceededException>(() => Parser.Parse(query, new ParserOptions { MaxDepth = 0 }));
+        }
+
+        [Fact]
+        public void Should_Throw_With_MaxDepth_1_On_SimpleQuery()
```
sungam3r, line 57: 👍

src/GraphQLParser/ParserContext.Parse.cs
sungam3r: In parser it is very important to produce and adhere to the general principles, otherwise it is very easy to get confused. What is the depth? When should it increase? Parser builds AST. We can say that the depth is the height of this AST. Sounds reasonable. So when to increment depth? The answer is - each time when new node is created by parser code (except in loops of cource). Even directive nodes can be deeply nested. Formally, even descriptions and comments should be counted for depth. These are also AST nodes.

sungam3r: So regarding your comment - I would move `IncreaseDepth()` calls from here and other places more closer to `NodeHelper.CreateGraphQLXXX` calls. Then it will be easy to understand where you need to put them and where you do not need (for now and in future).

Shane32: This PR has IncreaseDepth calls wherever there is a loop containing recursive calls. I’m not adding it throughout the whole library. What you say is true in principle so if you want to write a PR that achieves the same thing, go ahead.

sungam3r: See #157. I have doubts.

Shane32: > doubts

Yes, I certainly may have missed something. Your approach would leave less room for error. I think it’s covered though.

Shane32: I’m okay with depth being an arbitrary number that has a meaning only to the parser internals. Or with it having a logical value. The purpose isn’t for document validation - that’s part of GraphQL.NET. The purpose is preventing stack overflow exceptions which will crash the server. 

On this note I’d probably set the max depth to 1000 or some such large value. Except it just isn’t necessary, and both json serialization engines have a default max depth of 64. So to match it up I set it to 64.

sungam3r: Even 64 is quite enough.
`{ a { b { c } d { e { f } } g { h { i { k } } } } }` has max depth 13.

sungam3r: So each `{` increases depth by 2 .

Shane32: As of now each `{` only increases by 1, plus the base as 1.  Tests added to demonstrate.

sungam3r: I understand. It's logical from the point of view of a human. A lot of _intermediate_ levels are hidden from the point of view of the GraphQL grammar.

Shane32:
@sungam3r Is this PR ready to be merged?

src/GraphQLParser/ParserContext.Parse.cs
Shane32: Right. It was not specifically designed to be so; but that’s how the PR turned out, probably for the best.

sungam3r:
It conflicts (in terms of concepts) with #157. But probably we can merge and then remove everything not needed. OK! 👍

sungam3r (approved):

src/GraphQLParser/ParserContext.Parse.cs
sungam3r: >probably for the best

👍

Shane32:
Sure.  Although keep in mind this PR and the other one have completely different goals.

PR #157, opened 2022-01-03
Introduce CountVisitor and MaxDepthVisitor, tune depth calculation

sungam3r:
Relates to #153. @Shane32 _Depth_ term introduced here differs from one in #153. Here _depth_ (and _count) is built upon AST. Depth from #153 is built upon intuitive interpretation of source query. As you can see here depth (as well as count) is much more: simple source query `{a}` yields depth=5!
> Document->OperationDefinition->SelectionSet->Field->Name

Nevertheless, from the GraphQL grammar point of view, everything is correct. And if we take into account the correctness of this judgment, then #153 should work in the same terms. I confess, I don't like it very much, because for me it was not obvious at first that `{a}` had a depth of 5 :frowning:.


Shane32:
Is there a purpose of this visitor? Obviously it doesn’t prevent a stack overflow exception if it needs to parse the query first. I don’t mind this code but just curious what its purpose is. For a complexity analyzer?

sungam3r:
>  Obviously it doesn’t prevent a stack overflow exception

Not a goal.

Yes, it may be useful for complexity analyzer. You can also consider this code as a minimum instrumental set for AST analysis.

Shane32:
Perhaps we should unify the AST between the parser and GraphQL.NET before creating duplicate set of analysis tools.  As it is now I'm not even sure if this could be used by the complexity analyzer.  Do validation rules even have access to the AST as provided by the parser?  Or only the AST created/mapped within GraphQL.NET?

sungam3r:
This is a rather big change and I am not ready to speak about it.

src/GraphQLParser/Extensions/ASTNodeExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// Extension methods for <see cref="ASTNode"/>.
+    /// </summary>
+    public static class ASTNodeExtensions
+    {
+        /// <summary>
+        /// Counts the number of all nested nodes of the specified node along with that node itself.
+        /// </summary>
+        /// <param name="node">Parent node.</param>
+        /// <returns>The number of nodes.</returns>
+        public static int Count(this ASTNode node)
```
Shane32, line 16: We don't have any nodes with a Count property?  If so, this should be renamed so it does not conflict.  I feel also that this method name is not really descriptive of what it does.  The LINQ method called `Count()` simply counts direct descendants.  This counts all descendants.  We should have a more clear name here.  Perhaps `GetTotalCount()`.

src/GraphQLParser/Extensions/ASTNodeExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+            visitor.Visit(node, context).GetAwaiter().GetResult(); // it's safe since method is actually sync
+            return context.Count;
+        }
+
+        /// <summary>
+        /// Gets the maximum depth for this AST node.
+        /// </summary>
+        /// <param name="node">Parent node.</param>
+        /// <returns>Maximum depth. Minimum depth is 1.</returns>
+        public static int MaxDepth(this ASTNode node)
```
Shane32, line 29: Similarly, `MaxDepth()` doesn't seem descriptive enough.  And is max / maximum correct?  Seems to imply a variable or range, but the 'maximum depth' of a node has a fixed number, just like a count of descendants is a fixed number.  Should it be `GetTotalDepth()` ?  Just thinking out loud here.  But maybe `GetMaxDepth()` is fine.

src/GraphQLParser/Extensions/ASTNodeExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        public static int Count(this ASTNode node)
+        {
+            var visitor = new CountVisitor<DefaultCountContext>();
+            var context = new DefaultCountContext(_ => true);
+            visitor.Visit(node, context).GetAwaiter().GetResult(); // it's safe since method is actually sync
+            return context.Count;
+        }
+
+        /// <summary>
+        /// Gets the maximum depth for this AST node.
```
Shane32, line 25: We should probably add notes that say that this is the depth of the object hierarchy, etc.

Shane32 (approved):
Well, if you want to merge this, that's fine with me.  I just can't imagine how any code within GraphQL.NET can use it.

Attached are a few comments.  I am not sure we should include the extension methods at all, but it does not hurt.

Shane32 (changes requested):
See prior comments about naming extension methods

src/GraphQLParser/Extensions/ASTNodeExtensions.cs
sungam3r: > We don't have any nodes with a Count property?

No such nodes. Changed to `AllNestedCount`. Not a fan of _Get_ prefixes for extension methods.

src/GraphQLParser/Extensions/ASTNodeExtensions.cs
sungam3r: For me, _TotalDepth_ sounds even more incomprehensible. Changed to `MaxNestedDepth`.

src/GraphQLParser/Extensions/ASTNodeExtensions.cs
sungam3r: /// Gets the maximum depth of all nested nodes' chains for this AST node.

Shane32 (approved):
Not crazy about the name of the extension methods, but all good. :+1:

sungam3r:
I also tuned depth calculations.

src/GraphQLParser.Tests/ParserTests.cs
```diff
@@ -42,7 +42,7 @@ public void Should_Throw_With_Deep_Literal()
         [Fact]
         public void Should_Parse_With_Almost_Deep_Query()
         {
-            var count = 63;
+            var count = 30;
```
sungam3r, line 45: _Almost_Deep_Query_ seems strange now.

src/GraphQLParser.Tests/ParserTests.cs
```diff
@@ -87,31 +87,31 @@ public void Should_Throw_With_MaxDepth_0_On_SimpleQuery()
...
             var query = "{a}";
-            Should.Throw<GraphQLMaxDepthExceededException>(() => Parser.Parse(query, new ParserOptions { MaxDepth = 1 }));
+            Should.Throw<GraphQLMaxDepthExceededException>(() => Parser.Parse(query, new ParserOptions { MaxDepth = 4 }));
         }
 
         [Fact]
-        public void Should_Parse_With_MaxDepth_1_On_TypeDefinition()
+        public void Should_Parse_With_MaxDepth_3_On_TypeDefinition()
         {
             var query = "scalar Test";
```
sungam3r, line 106: Document ->ScalaTypeDefinition->Name

Shane32:
Well, those changes should have been a separate PR.  They have nothing to do with Countisitor and MaxDepthVisitor.

Shane32:
I think we should set the default depth to 128 if every 'logical' level is +2.  This will be similar, although not precisely, close to a JSON serializer's default level of 64.

sungam3r:
The changes go together to demonstrate new depth calculation and tests.

Shane32 (approved):

PR #159, opened 2022-01-03
New AST nodes

sungam3r:
Introduce new AST nodes:
- http://spec.graphql.org/October2021/#ArgumentsDefinition
- http://spec.graphql.org/October2021/#InputFieldsDefinition
- http://spec.graphql.org/October2021/#Arguments
- http://spec.graphql.org/October2021/#FieldsDefinition
- http://spec.graphql.org/October2021/#EnumValuesDefinition
- http://spec.graphql.org/October2021/#VariableDefinitions
- http://spec.graphql.org/October2021/#Directives
- http://spec.graphql.org/October2021/#ImplementsInterfaces

Goals:
- More strict grammar/parsing
- Less copy-paste in node visitors because of new virtual methods

### DIFF
--- a/src/GraphQLParser.Tests/ParserTests.cs
+++ b/src/GraphQLParser.Tests/ParserTests.cs
@@ -104,33 +104,33 @@ namespace GraphQLParser.Tests
             def.SelectionSet.Selections.Count.ShouldBe(1);
             var field = def.SelectionSet.Selections.First() as GraphQLField;
             field.SelectionSet.Selections.Count.ShouldBe(1);
-            field.Arguments.Count.ShouldBe(9);
+            field.Arguments.Items.Count.ShouldBe(9);
 
-            var boolValue = field.Arguments[0].Value.ShouldBeAssignableTo<GraphQLScalarValue>();
+            var boolValue = field.Arguments.Items[0].Value.ShouldBeAssignableTo<GraphQLScalarValue>();
             boolValue.Comment.ShouldNotBeNull().Text.ShouldBe("comment for bool");
 
-            var nullValue = field.Arguments[1].Value.ShouldBeAssignableTo<GraphQLScalarValue>();
+            var nullValue = field.Arguments.Items[1].Value.ShouldBeAssignableTo<GraphQLScalarValue>();
             nullValue.Comment.ShouldNotBeNull().Text.ShouldBe("comment for null");
 
-            var enumValue = field.Arguments[2].Value.ShouldBeAssignableTo<GraphQLScalarValue>();
+            var enumValue = field.Arguments.Items[2].Value.ShouldBeAssignableTo<GraphQLScalarValue>();
             enumValue.Comment.ShouldNotBeNull().Text.ShouldBe("comment for enum");
 
-            var listValue = field.Arguments[3].Value.ShouldBeAssignableTo<GraphQLListValue>();
+            var listValue = field.Arguments.Items[3].Value.ShouldBeAssignableTo<GraphQLListValue>();
             listValue.Comment.ShouldNotBeNull().Text.ShouldBe("comment for list");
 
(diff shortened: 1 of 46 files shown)


src/GraphQLParser/AST/ASTNodeKind.cs
```diff
@@ -125,7 +125,7 @@ public enum ASTNodeKind
         /// <summary>
         /// Boolean value. The two keywords true and false represent the two boolean values.
         /// <br/>
-        /// <see href="http://spec.graphql.org/October2021/#ObjectField"/>
+        /// <see href="http://spec.graphql.org/October2021/#BooleanValue"/>
```
sungam3r, line 128: oops!

src/GraphQLParser.Tests/ParserTests.cs
```diff
@@ -171,13 +171,13 @@ public void Comments_on_NamedTypes_Should_Read_Correctly(IgnoreOptions options)
             frag.TypeCondition.Type.Comment.Text.ShouldBe("comment for named type from TypeCondition");
 
             var def2 = document.Definitions[1] as GraphQLObjectTypeDefinition;
-            def2.Interfaces[0].Comment.Text.ShouldBe("comment for named type from ImplementsInterfaces");
+            def2.Interfaces.Items[0].Comment.Text.ShouldBe("comment for named type from ImplementsInterfaces");
```
Shane32, line 174: Is it necessary to change everything like this?  Can we not add an `Item[int index]` property to these classes for convenience?  Can they implement `IList`?

Do these additional 'list' classes hold information that they were unable to previously?

Shane32 (approved):

src/GraphQLParser.Tests/ParserTests.cs
Shane32: Looks like they hold can location information...

sungam3r: > Do these additional 'list' classes hold information that they were unable to previously?

Yes. The existence of these classes allows you to write virtual methods passing them in.

>  Can we not add an Item[int index] property to these classes for convenience? Can they implement IList?

OK. I thought to do it, but then I forgot. They all really encapsulate only one collection of one type so it can be done.

Shane32: I'm okay either way.  I would implement `IReadOnlyList<T>` at a minimum -- basically `IEnumerable`, `Item` and `Count` -- so that we can iterate over the collection or use Item and Count for allocation-free iteration.

Shane32: But whatever you think.

sungam3r: Done. Now I see only `VariableDefinitions`->`Variables` rename in tests.

sungam3r: Other properties in definitions classes do not have _Definitions_ suffix, so I deleted it here.

src/GraphQLParser/AST/ASTListNode.cs
```diff
@@ -0,0 +1,30 @@
+using System.Collections;
+using System.Collections.Generic;
+
+namespace GraphQLParser.AST
+{
+    /// <summary>
+    /// Represents a AST node that holds a list of other (nested) AST nodes.
+    /// </summary>
+    public abstract class ASTListNode<TNode> : ASTNode, IEnumerable<TNode>
```
Shane32, line 9:
```suggestion
    public abstract class ASTListNode<TNode> : ASTNode, IReadOnlyList<TNode>
```
No other changes needed here (but probably api approvals would change)

sungam3r: done

PR #161, opened 2022-01-04
Ensure that depth is not mismatched

Shane32:


src/GraphQLParser/ParserContext.Parse.cs
```diff
@@ -25,6 +26,9 @@ public GraphQLDocument ParseDocument()
             );
             _document.Definitions = definitions;
             _document.UnattachedComments = _unattachedComments;
+
+            Debug.Assert(_currentDepth == 1, "Depth has not returned to 1 after parsing document");
```
Shane32, line 30: This is not compiled within release builds, but will reliable fail tests if the number of `IncreaseDepth()` and `DecreaseDepth()` calls are mismatched.

sungam3r: 👍

Shane32:
@sungam3r There already is a bug in the code.  Trying to track it down.

src/GraphQLParser/ParserContext.Parse.cs
```diff
@@ -1327,7 +1331,6 @@ private GraphQLType ParseType()
                 return type;
             }
 
-            IncreaseDepth();
```
sungam3r, line 1330: In fact, this line should not be removed but I add needed calls in other places later.

sungam3r (approved):

PR #166, opened 2022-01-07
Rework of all value nodes + tune error messages

sungam3r:
I continue to work on strict grammar.


sungam3r:
I plan to add tests in further PRs to increase code coverage. Also I think about switching to file-scoped namespaces. There are no PRs in this repo so no conflicts.

src/GraphQLParser/AST/GraphQLScalarValue.cs
```diff
@@ -1,102 +0,0 @@
...
-    /// <see cref="ASTNodeKind.IntValue">Int</see>
-    /// <br/>
-    /// <see cref="ASTNodeKind.FloatValue">Float</see>
-    /// <br/>
-    /// <see cref="ASTNodeKind.EnumValue">Enumeration</see>
-    /// <br/>
-    /// <see cref="ASTNodeKind.NullValue">Null</see>
-    /// </summary>
-    [DebuggerDisplay("{Value}")]
-    public class GraphQLScalarValue : GraphQLValue
```
sungam3r, line 23: Bad design here, misleading naming. I removed this node. Also now all child nodes don't store `ASTNodeKind` -> less memory.

src/GraphQLParser.Tests/ParserTests.cs
```diff
@@ -527,6 +527,18 @@ public void Should_Throw_On_Unknown_OperationType()
             ex.Column.ShouldBe(1);
         }
 
+        [Theory]
+        [InlineData("enum E { true A }", "Unexpected Name \"true\"; enum values are represented as unquoted names but not 'true' or 'false' or 'null'.", 1, 10)]
+        [InlineData("enum E { B false }", "Unexpected Name \"false\"; enum values are represented as unquoted names but not 'true' or 'false' or 'null'.", 1, 12)]
+        [InlineData("enum E { A null B }", "Unexpected Name \"null\"; enum values are represented as unquoted names but not 'true' or 'false' or 'null'.", 1, 12)]
+        public void Should_Throw_On_Invalid_EnumValue(string query, string description, int line, int column)
```
sungam3r, line 534: Test failed before changes since EnumValueDefinition accepts true/false/null that contradicts spec.

Shane32 (approved):
I'll have to see the corresponding changes required in GraphQL.NET to really know how these changes will work.  I suggest having the dev branch pull in the beta build from GitHub Packages before publishing the new version here.  Here's a few concerns I see:

- Is there a unified way of retrieving the Value of a node -- or certainly a scalar node -- without having to cast it to it's specific node type?  Does it matter?
- How does elimination of GraphQLScalarValue from the inheritance chain affect use?

sungam3r:
I think that nothing bad will happen. See `CoreToVanillaConverter` especilly `internal static IValue Value(GraphQLValue source)` with switch by `ASTNodeKind`. I hope that we **can** eliminate `GraphQL.Language.AST` subsystem at all in v5. At least I have suspicion. The missing API can be added to the parser project.

sungam3r:
I would say that the parser object model / API evolved to such an extent that ready to "absorb" `GraphQL.Language.AST` (or be merged with it). Just open two projects side-by-side and compare file names, interfaces, API, etc.

Shane32:
Again, let's use the beta build in graphql.net before releasing vNext of parser, so we can make tweaks if necessary

sungam3r:
Of course. I was not going to release the parser separately.

sungam3r:
I still have work in the parser, a few PRs. When done I'll bump parser version on `develop` branch here.

PR #176, opened 2022-01-09
Add tests

sungam3r:


src/GraphQLParser/AST/GraphQLFloatValue.cs
```diff
@@ -1,12 +1,12 @@
...
 
 /// <summary>
-/// AST node for <see cref="ASTNodeKind.BooleanValue"/>.
+/// AST node for <see cref="ASTNodeKind.FloatValue"/>.
 /// </summary>
 public class GraphQLFloatValue : GraphQLValue
 {
     /// <inheritdoc/>
-    public override ASTNodeKind Kind => ASTNodeKind.BooleanValue;
+    public override ASTNodeKind Kind => ASTNodeKind.FloatValue;
```
sungam3r, line 9: OOOOPS

src/GraphQLParser/ParserContext.Parse.cs
```diff
@@ -1371,29 +1360,21 @@ private GraphQLType ParseType()
     // http://spec.graphql.org/October2021/#TypeExtension
     private GraphQLTypeExtension ParseTypeExtension()
     {
-        ExpectKeyword("extend");
```
sungam3r, line 1374: Bug here: _extend_ should be a part of _TypeExtensionDefinition_ . Location was incorrectly calculated.

PR #180, opened 2022-01-10
Add tests

sungam3r:


sungam3r:
https://app.codecov.io/gh/graphql-dotnet/parser
изображение

src/GraphQLParser/AST/GraphQLLocation.cs
```diff
@@ -1,12 +1,10 @@
 using System;
-using System.Diagnostics;
 
 namespace GraphQLParser.AST;
 
 /// <summary>
 /// Provides information regarding the location of a node within a document's original text.
 /// </summary>
-[DebuggerDisplay("(Start={Start}, End={End})")]
```
sungam3r, line 9: Not needed, `ToString()` exists.

sungam3r:
изображение

Shane32:
Nice!

sungam3r:
I doubt that some areas can be covered with tests. Like this one
изображение

Shane32:
It doesn’t count consistently either. You might find it counts it as 100%. But the hardest part is writing tests for error checks on private methods (which shouldn’t occur) or values returned by external methods which should never return an error. Mocks help but can’t always simulate the issue if the external method is part of a sealed class.

Shane32:
For my recent small projects built with 100% testing in mind, eventually it seems a bit pointless as you get much above 90%. On my corporate code where I recently started a new project “tests-first” I’m aiming for at least 80%. But aiming for a higher percentage increasingly becomes less productive with little value.

Shane32:
Like all the serializers work I’m doing, I could simulate invalid json syntax for every type being deserialized, hitting every bit of “throw” code…. But I can’t simulate the json parser returning a value node where it expects a property node.

Shane32:
It is really satisfying to see the coverage at 98%+ though!

sungam3r:
As I was convinced, it is important not so much the value of the coverage, how much the process of finding errors while increasing this coverage.

Shane32:
For sure. I’ve found so many bugs while attempting to increase coverage.

PR #184, opened 2022-01-13
GraphQLName implicit cast to ROM

sungam3r:


sungam3r:
@Shane32 I'm working on better parser integration into the main project.

sungam3r:
I merge it and go on with other changes.

Shane32:
Not sure why the implicit cast to ROM is here, but maybe add an implicit or explicit cast to string also?

sungam3r:
OK

sungam3r:
Implicit cast to ROM for
```
public void DoSome(Rom a) { }

var name = fragment.Name; // (GraphQLName type)
DoSome(name); // vs DoSome(name.Value) everywhere
```

PR #186, opened 2022-01-14
ROM optimizations

sungam3r:
These changes will allow us to work with ROM more widely/freely in the main project. Many places that use `string` can be changed to `ROM` without performance drop (I think).

### DIFF
--- a/src/GraphQLParser.Tests/GraphQLParser.Tests.csproj
+++ b/src/GraphQLParser.Tests/GraphQLParser.Tests.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6;net5;netcoreapp3.1</TargetFrameworks>
+    <TargetFramework>net6</TargetFramework>
     <Nullable>disable</Nullable>
     <NoWarn>$(NoWarn);1591;IDE0008;IDE0022;IDE0058;IDE1006</NoWarn>
   </PropertyGroup>
--- a/src/GraphQLParser/AST/GraphQLName.cs
+++ b/src/GraphQLParser/AST/GraphQLName.cs
@@ -12,6 +12,21 @@ public class GraphQLName : ASTNode
     public override ASTNodeKind Kind => ASTNodeKind.Name;
 
     /// <summary>
+    /// Creates a new instance (with empty value).
+    /// </summary>
+    public GraphQLName()
+    {
+    }
+
+    /// <summary>
+    /// Creates a new instance with the specified value.
+    /// </summary>
+    public GraphQLName(ROM value)
+    {
+        Value = value;
+    }
+
+    /// <summary>
     /// Name value represented as <see cref="ROM"/>.
     /// </summary>
     public ROM Value { get; set; }
--- a/src/GraphQLParser/ROM.cs
+++ b/src/GraphQLParser/ROM.cs
@@ -53,10 +53,28 @@ public readonly struct ROM : IEquatable<ROM>
         if (_memory.Equals(other._memory))
             return true;
 
-        // then check byte by byte
-        if (_memory.Length != other._memory.Length)
-            return false;
+        // then check byte by byte, SequenceEqual already has length check
+        //[MethodImpl(MethodImplOptions.AggressiveInlining)]
+        //public static bool SequenceEqual<T>(this ReadOnlySpan<T> span, ReadOnlySpan<T> other) where T : IEquatable<T>
+        //{
+        //    int length = span.Length;
+        //    if (default(T) != null && IsTypeComparableAsBytes<T>(out NUInt size))
+        //    {
+        //        if (length == other.Length)
+        //        {
+        //            return SpanHelpers.SequenceEqual(ref Unsafe.As<T, byte>(ref MemoryMarshal.GetReference(span)), ref Unsafe.As<T, byte>(ref MemoryMarshal.GetReference(other)), (NUInt)length * size);
+        //        }
 
+        //        return false;
+        //    }
+
+        //    if (length == other.Length)
+        //    {
+        //        return SpanHelpers.SequenceEqual(ref MemoryMarshal.GetReference(span), ref MemoryMarshal.GetReference(other), length);
+        //    }
+
+        //    return false;
+        //}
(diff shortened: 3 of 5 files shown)


src/GraphQLParser/ROM.cs
```diff
@@ -40,39 +40,28 @@ public ROM(ReadOnlyMemory<char> memory)
...
-            int num4 = span[i + 1];
-            //if (num4 != 0)
-            num2 = (num2 << 5) + num2 ^ num4;
-            //else
-            //    break;
-        }
-
-        return num1 + num2 * 1566083941;
-    }
+    public override int GetHashCode() => _memory.GetHashCode();
```
sungam3r, line 64: @Shane32 My fault here :( 

изображение

src/GraphQLParser.Tests/GraphQLParser.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6;net5;netcoreapp3.1</TargetFrameworks>
+    <TargetFramework>net6</TargetFramework>
```
sungam3r, line 4: net6 only, OK?

src/GraphQLParser/ROM.cs
Shane32: How is the hash code used? Is .NET’s hash code implementation based off of the underlying data or the memory address of the data? Does it matter?

Shane32: For structs .NET should already gather underlying members to use for the hash code. Is this method necessary? It should be identical to what is default. Is it faster? Or just here to be explicit?

src/GraphQLParser/ROM.cs
```diff
@@ -40,39 +40,46 @@ public ROM(ReadOnlyMemory<char> memory)
...
+        //    }
+
+        //    if (length == other.Length)
+        //    {
+        //        return SpanHelpers.SequenceEqual(ref MemoryMarshal.GetReference(span), ref MemoryMarshal.GetReference(other), length);
+        //    }
+
+        //    return false;
+        //}
         return _memory.Span.SequenceEqual(other._memory.Span);
```
Shane32, line 78: Might want to check if this method is faster for .NET Standard 2.1:

https://docs.microsoft.com/en-us/dotnet/api/system.memoryextensions.equals?view=net-6.0#system-memoryextensions-equals(system-readonlyspan((system-char))-system-readonlyspan((system-char))-system-stringcomparison)

src/GraphQLParser/ROM.cs
sungam3r: > How is the hash code used? Is .NET’s hash code implementation based off of the underlying data or the memory address of the data? Does it matter?

I overrided equals so compiler wanted me to override GetHashCode to get rid of warning.

> For structs .NET should already gather underlying members to use for the hash code. Is this method necessary? It should be identical to what is default. Is it faster? Or just here to be explicit?

Also it's standard practice to override GetHasCode for structs. Default implementation works via reflection.

sungam3r: This method is not used anywhere (explicitly).

sungam3r: > Is it faster?

Yes.

src/GraphQLParser/ROM.cs
sungam3r: I did not check. It has already been used. I only added an additional call above which should work pretty quickly.

src/GraphQLParser.Tests/GraphQLParser.Tests.csproj
Shane32: Sure

Shane32 (approved):
See note about suggested alternative rom comparison method

src/GraphQLParser/ROM.cs
Shane32: Well if you are planning to use this call extensively, you might want to.

src/GraphQLParser/AST/GraphQLName.cs
```diff
@@ -11,6 +11,21 @@ public class GraphQLName : ASTNode
...
+    /// Creates a new instance (with empty value).
+    /// </summary>
+    public GraphQLName()
+    {
+    }
+
+    /// <summary>
+    /// Creates a new instance with the specified value.
+    /// </summary>
+    public GraphQLName(ROM value)
```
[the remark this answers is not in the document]
Shane32: > These changes will allow us to work with ROM more widely/freely in the main project. Many places that use `string` can be changed to `ROM` without performance drop (I think).

This will need to be verified.  If it is stored as ROM in the main project, and then converted to a string more than once, it is a guaranteed performance drop.  If it's stored as a ROM and converted to a string once, then it's the same.  Only if it's stored as a ROM and unused would there be a performance benefit.  I already tried using ROMs throughout the main project and it did not go well.  I certainly would not use ROMs in the main project 'just because we can', so if there is no performance difference, we should leave it as strings.

src/GraphQLParser/ROM.cs
sungam3r: Oh, it's another method. OK, added TODO to check later.

src/GraphQLParser/AST/GraphQLName.cs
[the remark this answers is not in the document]
sungam3r: ROM -> string is explicit conversion (as designed) as it _may_ allocate string.
string -> ROM is implicit as it never allocate.

I added code to compare ROM vs ROM when ROM _may be_ backed by string. 

> Only if it's stored as a ROM and unused would there be a performance benefit. I already tried using ROMs throughout the main project and it did not go well. I certainly would not use ROMs in the main project 'just because we can', so if there is no performance difference, we should leave it as strings.

Will see.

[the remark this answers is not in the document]
sungam3r: I still do not know what happens.

src/GraphQLParser/ROM.cs
sungam3r: @Shane32 See #190. This is a rather interesting case for me.

PR #190, opened 2022-01-14
GetHashCode fix

sungam3r:


src/GraphQLParser/ROM.cs
```diff
@@ -95,7 +95,27 @@ public static bool IsEmptyOrWhiteSpace(ROM value)
...
+        //     return CombineHashCodes(_object.GetHashCode(), _index.GetHashCode(), _length.GetHashCode());
+        // }
+
+        //return _memory.GetHashCode();
+
+        // TODO: think about GetHashCode implementation
+        if (_memory.Length == 0)
+            return 0;
+        var span = _memory.Span;
+        return span[0] ^ span[span.Length - 1];
```
Shane32, line 117:
```suggestion
        return string.GetHashCode(_memory.Span);
```
How about this?

Shane32: Only available in .NET 3.0+, but I expect that it's extremely optimized.

Shane32: I'd suggest a TFM ifdef check and use this when possible, otherwise the old code that computed the value.

sungam3r: Probably. But the problem is the _implementation_ itself!

Shane32: Whatever you think though. :+1:

sungam3r: I need such a behavior in which the hash will be the same if memory points to the same data regardless of the backed object.

sungam3r: // ReadOnlyMemory<T> implementation
        // public override int GetHashCode()
        // {
        //     if (_object == null)
        //     {
        //         return 0;
        //     }
        //
        //     return CombineHashCodes(_object.GetHashCode(), _index.GetHashCode(), _length.GetHashCode());
        // }

I need to see the source code of the proposed method, but most likely it is unmanaged.

sungam3r: I will explain what's the problem - if for 2 objects the hash code is different, then many container classes (Dictionary, HashSet, etc) consider objects **different**.

sungam3r:
I merge this on and go on.

src/GraphQLParser/ROM.cs
Shane32: > I need such a behavior in which the hash will be the same if memory points to the same data regardless of the backed object.

`string.GetHashCode` does exactly that.

Shane32: at least, i would be very much surprised if not.

sungam3r: > string.GetHashCode does exactly that.

Good if so.

PR #192, opened 2022-01-14
Add GraphQLName.ToString/StringValue

sungam3r:
A lot of cases use `GraphQLName` in printing error messages:
изображение

`ToString` is called during string interpolation.

### DIFF
--- a/src/GraphQLParser/AST/GraphQLName.cs
+++ b/src/GraphQLParser/AST/GraphQLName.cs
@@ -8,6 +8,9 @@ namespace GraphQLParser.AST;
 [DebuggerDisplay("GraphQLName: {Value}")]
 public class GraphQLName : ASTNode
 {
+    private ROM _value;
+    private string? _string;
+
     /// <inheritdoc/>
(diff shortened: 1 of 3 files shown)


sungam3r:
`StringValue` is added for convenience.

sungam3r:
изображение

"Only" 160 left :)

src/GraphQLParser/AST/GraphQLName.cs
```diff
@@ -26,10 +29,42 @@ public GraphQLName(ROM value)
...
+            _value = value;
+            _string = null;
+        }
+    }
+
+    /// <summary>
+    /// Name value represented as <see cref="string"/>.
+    /// <br/>
+    /// This property allocates the string on the heap on first access
+    /// and then caches it until <see cref="Value"/> does not change.
```
Shane32, line 50: “and then caches it as long as…”

sungam3r: 👍

sungam3r: done in subsequent PR

PR #193, opened 2022-01-15
Refactoring of value types

sungam3r:
Final preparations to changes in the main repo.

Goal: get rid of `CoreToVanillaConverter` and all bits from `GraphQL.Language.AST` namespace
Idea: we don't need converting parser AST into GraphQL.NET AST since parser API now has all necessary tools.

UPDATES:
1. Added: IHasValueNode interface
2. Added: ctors with CLR objects AND with ROM for "primitive" value nodes + for comment and description
3. Removed: GraphQLListValue.Value
4. Added: GraphQLName operators


src/GraphQLParser.Tests/LexerTests.cs
```diff
@@ -1055,174 +453,4 @@ public void Lex_BlockString_Tests(int number, string input, string expected)
...
-    {
-        return ")".Lex();
-    }
-
-    private static Token GetSimpleStringTokenLexer()
-    {
-        return "\"str\"".Lex();
-    }
-
-    private static Token GetSingleDecimalIntTokenLexer()
```
sungam3r, line 1139: I merge each group of 4 tests (kind, start, end, value) into one test.

src/GraphQLParser.Tests/LexerTests.cs
```diff
@@ -362,657 +159,258 @@ public void Lex_VeryLongStringToken(bool escape)
...
+        var token = ",,,foo,,,".Lex();
+        token.Kind.ShouldBe(TokenKind.NAME);
+        token.Start.ShouldBe(3);
+        token.End.ShouldBe(6);
+        token.Value.ShouldBe("foo");
     }
 
     [Fact]
-    public void Lex_RightBracketPunctuation_HasCorrectStart()
+    public void NameWithHyphen()
```
sungam3r, line 320: One new test.

src/GraphQLParser/AST/GraphQLName.cs
```diff
@@ -47,7 +46,7 @@ public ROM Value
     /// Name value represented as <see cref="string"/>.
     /// <br/>
     /// This property allocates the string on the heap on first access
-    /// and then caches it until <see cref="Value"/> does not change.
+    /// and then caches it as long as <see cref="Value"/> does not change.
```
sungam3r, line 49: fixed

Shane32:
There’s a number of changes in the last few pull requests that cache clr values and strings. We need to take into account the increased memory use here versus before these cached values were added when considering any benefits attained in the main repo.

src/GraphQLParser/AST/Values/GraphQLValue.cs
```diff
@@ -7,4 +7,8 @@ namespace GraphQLParser.AST;
 /// </summary>
 public abstract class GraphQLValue : ASTNode
 {
+    /// <summary>
+    /// Returns the CLR object that represents the value of this AST node.
+    /// </summary>
+    public abstract object? ClrValue { get; }
```
sungam3r, line 13: All value nodes now can return CLR representation of the parsed value. Some nodes have backing fields, some not (NullValue, BooleanValue, EnumValue).

src/GraphQLParser/Shim/Numbers.cs
```diff
@@ -0,0 +1,80 @@
+using System;
+using System.Globalization;
+
+namespace GraphQLParser;
+
+/// <summary>
+/// Adapter to unify usages of int.TryParse(ReadOnlySpan) for netstandard2.0 and netstandard2.1
+/// </summary>
+internal static class Int
```
sungam3r, line 9: Copied without changes from the main repo.

src/GraphQLParser/Shim/DecimalData.cs
```diff
@@ -0,0 +1,29 @@
...
+// https://github.com/microsoft/referencesource/blob/master/mscorlib/system/decimal.cs
+// The lo, mid, hi, and flags fields contain the representation of the
+// Decimal value. The lo, mid, and hi fields contain the 96-bit integer
+// part of the Decimal. Bits 0-15 (the lower word) of the flags field are
+// unused and must be zero; bits 16-23 contain must contain a value between
+// 0 and 28, indicating the power of 10 to divide the 96-bit integer part
+// by to produce the Decimal value; bits 24-30 are unused and must be zero;
+// and finally bit 31 indicates the sign of the Decimal value, 0 meaning
+// positive and 1 meaning negative.
+internal readonly struct DecimalData
```
sungam3r, line 13: Copied without changes from the main repo.

src/GraphQLParser/ParserContext.Parse.cs
```diff
@@ -999,6 +999,7 @@ private GraphQLValue ParseObjectValue(bool isConstant)
 
         val.Comment = GetComment();
         val.Fields = ZeroOrMore(TokenKind.BRACE_L, isConstant ? constant : value, TokenKind.BRACE_R);
+        val.Value = _source.Slice(start, _currentToken.End - start - 1);
```
sungam3r, line 1002: Small fix here - `GraphQLListValue` had `Value` property, but `GraphQLObjectValue` did not.

src/GraphQLParser/AST/GraphQLIntValue.cs
```diff
@@ -8,13 +11,72 @@ namespace GraphQLParser.AST;
...
+        }
+    }
+
+    /// <summary>
+    /// Integer value represented as <see cref="int"/>, <see cref="long"/>, <see cref="decimal"/> or <see cref="BigInteger"/>.
+    /// <br/>
+    /// This property allocates the string on the heap on first access
+    /// and then caches it as long as <see cref="Value"/> does not change.
+    /// </summary>
+    private /*public*/ object TypedValue //TODO: ??? no typed value :(
```
sungam3r, line 39: @Shane32 Please pay attention to `IntValue` and `FloatValue`. I did so, because the other options were even worse.

sungam3r: `Parse` methods were ported from the main repo.

sungam3r:
I understand. These cached values are not always used. Winning is achieved through the complete elimination of the conversion of one AST to another. The number of objects created during conversion is more than the number of strings created for AST **value** nodes (note that only value nodes).

sungam3r:
I specifically marked all the places in the main repo that allocate strings and work on the elimination of these allocations. While everything goes according to plan. Somewhere without any problems, it is possible to avoid allocations completely.

src/GraphQLParser/AST/GraphQLIntValue.cs
```diff
@@ -8,13 +11,72 @@ namespace GraphQLParser.AST;
...
+                }
+
+                // Since BigInteger can contain any valid integer (arbitrarily large), this is impossible to trigger via an invalid query
+                throw new InvalidOperationException($"Invalid number {Value}");
+            }
+        }
+    }
+
+    /// <inheritdoc />
+    public override object? ClrValue => TypedValue;
```
Shane32, line 79:
```suggestion
    public override object? ClrValue => _number ??= TypedValue;
```

Shane32: This should inline

src/GraphQLParser/AST/GraphQLIntValue.cs
```diff
@@ -8,13 +11,72 @@ namespace GraphQLParser.AST;
...
+    private /*public*/ object TypedValue //TODO: ??? no typed value :(
+    {
+        get
+        {
+            return _value.Length == 0
+                ? throw new InvalidOperationException("Invalid number (empty string)")
+                : _number ??= Parse();
+
+            object Parse()
+            {
```
Shane32, line 48:
```suggestion
            if (_value.Length == 0)
                throw new InvalidOperationException("Invalid number (empty string)");

```

src/GraphQLParser/AST/GraphQLIntValue.cs
```diff
@@ -8,13 +11,72 @@ namespace GraphQLParser.AST;
...
+
+                // If the value doesn't fit in an decimal, revert to using BigInteger...
+                if (BigInt.TryParse(Value, NumberStyles.AllowLeadingSign, CultureInfo.InvariantCulture, out var bigIntegerResult))
+                {
+                    return bigIntegerResult;
+                }
+
+                // Since BigInteger can contain any valid integer (arbitrarily large), this is impossible to trigger via an invalid query
+                throw new InvalidOperationException($"Invalid number {Value}");
+            }
```
Shane32, line 74:
```suggestion
```
Needs reformatting also of course.

src/GraphQLParser/AST/GraphQLFloatValue.cs
```diff
@@ -8,13 +10,74 @@ namespace GraphQLParser.AST;
...
+    private /*public*/ object TypedValue //TODO: ??? no typed value :(
+    {
+        get
+        {
+            return _value.Length == 0
+                ? throw new InvalidOperationException("Invalid number (empty string)")
+                : _number ??= Parse();
+
+            object Parse()
+            {
```
Shane32, line 47:
```suggestion
            if (_value.Length == 0)
                throw new InvalidOperationException("Invalid number (empty string)");

```

src/GraphQLParser/AST/GraphQLFloatValue.cs
```diff
@@ -8,13 +10,74 @@ namespace GraphQLParser.AST;
...
+                    // Cast the decimal to our struct to avoid the decimal.GetBits allocations.
+                    var decBits = System.Runtime.CompilerServices.Unsafe.As<decimal, DecimalData>(ref dec);
+                    decimal temp = new decimal(dbl);
+                    var dblAsDecBits = System.Runtime.CompilerServices.Unsafe.As<decimal, DecimalData>(ref temp);
+                    if (!decBits.Equals(dblAsDecBits))
+                        return dec;
+                }
+
+                return dbl;
+            }
```
Shane32, line 75:
```suggestion
            return dbl;
```

Shane32: Needs reformatting

src/GraphQLParser/AST/GraphQLFloatValue.cs
```diff
@@ -8,13 +10,74 @@ namespace GraphQLParser.AST;
...
+        }
+    }
+
+    /// <inheritdoc />
+    public override object? ClrValue => TypedValue;
```
Shane32, line 80:
```suggestion
    public override object? ClrValue => _number ??= TypedValue;
```
Should inline

src/GraphQLParser/AST/GraphQLObjectValue.cs
```diff
@@ -7,10 +8,58 @@ namespace GraphQLParser.AST;
...
+                    @object.Add(field.Name.StringValue, field.Value.ClrValue);
+                _object = @object;
+            }
+
+            return _object;
+        }
+    }
+
+    /// <inheritdoc />
+    public override object? ClrValue => TypedValue;
```
Shane32, line 62:
```suggestion
    public override object? ClrValue => _object ??= TypedValue;
```
(needs other changes here)

And similar changes throughout, so it inlines.

Shane32 (changes requested):
I would change it so `ClrValue` returns `_value ??= TypedValue` throughout, so it can be inlined.

src/GraphQLParser/AST/GraphQLIntValue.cs
sungam3r: 👍

Shane32:
In general, this design is starting to be at odds with itself as a standalone parser.  I'm not complaining, just pointing out that now we are mixing the code that interprets the values into the parser.  When do we stop?  Keep in mind that each ROM hold a pointer to the exact same originally-parsed array and consumes memory for the array pointer, offset and length.  For example, if every single string is eventually converted from a `ROM` to a `string`, and if we are holding a cached copy of the string, then holding the ROM is pointless.  We can just return a ROM created from the string if the ROM was ever requested, and there is no heap allocation for the conversion.  Likewise, it would save memory for a boolean value to only ever hold a boolean, and requesting a ROM from it would return a static ROM just like we do for boxed booleans.

Bottom line, if we don't use the ROMs outside of this project, then holding the ROMs become a waste of memory, and it is better off to just deserialize the data right away.

Yes there are a few instances where it could be faster to interpret it dynamically based on need.  These are going to be limited to value literals (int, float, string), comments, and that's about it.  Query validation examines each node so even if there was a request with unused operations, it is likely that every node is going to be parsed in some fashion.

src/GraphQLParser/AST/GraphQLObjectValue.cs
sungam3r: done

sungam3r:
It is important to see a complete picture. These changes **can** add some memory, but it is important that they allow you to get rid of **much** more memory losses in the main project. I'm working on it.

sungam3r:
I advise you to look at `CoreToVanillaConverter` and think why it is necessary. What does this class do? Why?

sungam3r:
> In general, this design is starting to be at odds with itself as a standalone parser. 

I know. This is a compromise to achive the best end-to-end results.

src/GraphQLParser/AST/Values/GraphQLValue.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// Returns the CLR object that represents the value of this AST node.
+    /// <br/>
+    /// This value is cached. Call <see cref="Reset"/> to reset it when needed.
+    /// </summary>
+    public abstract object? ClrValue { get; }
+
+    /// <summary>
+    /// Resets cached <see cref="ClrValue"/>.
+    /// </summary>
+    public abstract void Reset();
```
sungam3r, line 20: @Shane32 I added explicit `Reset`. It's simpler. In practice, it is unlikely that it will be necessary to someone, but if the value is cached, I was obliged to provide a reset method.

src/GraphQLParser/ParserContext.Parse.cs
```diff
@@ -855,7 +855,6 @@ private GraphQLValue ParseListValue(bool isConstant)
 
         val.Comment = GetComment();
         val.Values = ZeroOrMore(TokenKind.BRACKET_L, isConstant ? constant : value, TokenKind.BRACKET_R);
-        val.Value = _source.Slice(start, _currentToken.End - start - 1);
```
sungam3r, line 858: `Value` property for both `ListValue` and `ObjectValue` was a mistake. These are composite objects that do not have their own data like any other "top-level" AST nodes. Only "primitive" AST nodes should have `ROM Value { get; set; }` property.

src/GraphQLParser/AST/Values/GraphQLValue.cs
sungam3r: +NameNode of course (not a value node)

sungam3r:
@Shane32 100% diff coverage, I think I finished here and Pr is ready for review. There are a lot of boring work in the main repository changing scalar parsing.

Shane32:
Merge if you like. Or I can review later tonight. But I think I understand generally what this PR is.

src/GraphQLParser/AST/GraphQLName.cs
```diff
@@ -29,35 +28,27 @@ public GraphQLName(ROM value)
...
+    public ROM Value { get; set; }
 
     /// <summary>
     /// Name value represented as <see cref="string"/>.
     /// <br/>
     /// This property allocates the string on the heap on first access
-    /// and then caches it until <see cref="Value"/> does not change.
+    /// and then caches it. Call <see cref="Reset"/> to reset cache when needed.
     /// </summary>
     public string StringValue
```
sungam3r, line 42: @Shane32 Change to `TypedValue`? Some value nodes have `TypedValue` property. `GraphQLName` is not a value node though.

Example of changes:
изображение

sungam3r:
I'll wait. I'm interested in review. I use local builds to reference from the main project. Also note that GraphQL.NET has much more changes than here. I do not remember when the last time I made such edits if ever.

sungam3r:
If you want, I can publish PR in GraphQL.NET for you to understand what I'm doing. ~150 tests failed for now.

Shane32:
> If you want, I can publish PR in GraphQL.NET for you to understand what I'm doing. ~150 tests failed for now.

Good idea.

Shane32:
I see there's a number of classes where you have added a `Reset` method and removed the set method.  I fail to understand why you would add the `Reset` method.  It seems counterintuitive.  You seemed to have a good design before, where setting the rom would reset the cache.  Makes sense; why change it?  Now you have to set ROM and call Reset?  I don't see any benefit here, just an added complexity.

I think perhaps these questions should be answered:
- Why isn't there a constructor accepting a ROM?
- Why is there a blank constructor?
- When in practice is ROM set excluding during initialization?
- Why can you set the value at all?

It seems to me that there should be a constructor for these classes that accept a ROM, and `Reset` should be removed, and the set method restored to as it was prior to this PR.

Why don't we make all the classes read-only?  Perhaps utilize the `record` keyword if it makes sense also; I'm not sure.

sungam3r:
> Now you have to set ROM and call Reset?

No. No one ever needs to reset cache. Even I.

> Makes sense; why change it? 

Because complex value nodes have no ROM property but have other properties (fields for object value node) that **may** be changed. I don't want to bother with all that property change tracking. Again - in real life no one needs it. So I removed tracking and added explicit methods. 

> Why isn't there a constructor accepting a ROM?

I can add one everywhere though it's not nessesary. 

> Why is there a blank constructor?

For convenience. To be able to create node first. In NodeHelper for example.

> When in practice is ROM set excluding during initialization?

Now nowhere. I doubt if anywhere else in future but who knows.

> Why can you set the value at all?

I don't understand.

> Why don't we make all the classes read-only?

I would not. It will disallow changing AST by node visitors. It's rather powerful feature.

Shane32:
Ok

Shane32:
Ok, but I don't like the answer.  The reason stated for allowing changing values at all is "changing AST by node visitors."  This means it is designed to be used by user code.  Users shouldn't be expected to have to call Reset.

Personally I would add the change tracking back for all of the easy ones, and if you need Reset for ObjectValueNode, so be it.  I guess it doesn't matter too much at the moment.

sungam3r:
> for all of the easy ones

I did not want to selectively make decisions ("easy" value node or not "easy" value node). There are nodes with values. Values are somehow materialized to objects. These objects can be stored, they can be changed. Now the caching/reset is simple and understandable. It is unlikely that someone will need to change them at all.

sungam3r:
Of course, this is a compromise. At first I tried to inherit from base value nodes. It did not work.

Shane32:
I'm not happy we are boxing all values. We can fix it, but it pushes the problem on to `IntGraphType` and similar, which would just box it there.  So might as well box them here and cache them as boxed values.

Shane32:
We should add some simple code to `GraphQLIntValue` to use pre-boxed values for -1 to 10, or some such range.  At least 0 to 5.

Shane32:
Question: why do we add all these constructors instead of just having `ClrValue` as a writable property?  We could even remove `Reset`.

```
    public class GraphQLIntValue : GraphQLParser.AST.GraphQLValue
    {
        public GraphQLIntValue() { }
        public GraphQLIntValue(byte value) { }
        public GraphQLIntValue(decimal value) { }
        public GraphQLIntValue(short value) { }
        public GraphQLIntValue(int value) { }
        public GraphQLIntValue(long value) { }
        public GraphQLIntValue(System.Numerics.BigInteger value) { }
        public GraphQLIntValue(sbyte value) { }
        public GraphQLIntValue(ushort value) { }
        public GraphQLIntValue(uint value) { }
        public GraphQLIntValue(ulong value) { }
        public override object? ClrValue { get; }
        public GraphQLParser.ROM Value { get; set; }
        public override void Reset() { }
    }```

vs

```
    public class GraphQLIntValue : GraphQLParser.AST.GraphQLValue
    {
        public GraphQLIntValue() { }
        public override object? ClrValue { get; set; }
        public GraphQLParser.ROM Value { get; set; }
    }
```

File: src/GraphQLParser/AST/Values/GraphQLListValue.cs [hunk 19]
```diff
@@ -9,18 +10,53 @@ namespace GraphQLParser.AST;
...
+                if (Values == null || Values.Count == 0)
+                {
+                    _list = Array.Empty<object?>();
+                }
+                else
+                {
+                    var list = new List<object?>(Values.Count);
+                    foreach (var value in Values)
+                        list.Add(value.ClrValue);
+                    _list = list;
```
Shane32, line 44: Count is known; this should be an array to eliminate an allocation, especially as arrays implement `IList`.

src/GraphQLParser/AST/Values/GraphQLListValue.cs
```diff
@@ -9,18 +10,53 @@ namespace GraphQLParser.AST;
...
 
     /// <summary>
-    /// Values of the list represented as a list of nested <see cref="GraphQLValue"/> nodes.
+    /// List value represented as <see cref="List{T}"/>.
+    /// <br/>
+    /// This property allocates the string on the heap on first access
+    /// and then caches it. Call <see cref="Reset"/> to reset cache when needed.
     /// </summary>
-    public List<GraphQLValue>? Values { get; set; }
+    public IList<object?> TypedValue
```
Shane32, line 29: Why public here and nowhere else?

src/GraphQLParser/AST/Values/GraphQLObjectValue.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// </summary>
+    public List<GraphQLObjectField>? Fields { get; set; }
+
+    /// <summary>
+    /// Object value represented as <see cref="IDictionary{TKey, TValue}"/>.
+    /// <br/>
+    /// This property allocates the string on the heap on first access
+    /// and then caches it. Call <see cref="Reset"/> to reset cache when needed.
+    /// </summary>
+    public IDictionary<string, object?> TypedValue
```
Shane32, line 28: Also here

src/GraphQLParser.ApiTests/GraphQLParser.approved.txt
```diff
@@ -469,8 +499,12 @@ namespace GraphQLParser.AST
     public class GraphQLStringValue : GraphQLParser.AST.GraphQLValue
     {
         public GraphQLStringValue() { }
+        public GraphQLStringValue(string value) { }
+        public override object? ClrValue { get; }
         public override GraphQLParser.AST.ASTNodeKind Kind { get; }
+        public string TypedValue { get; }
```
Shane32, line 505: Even here `TypedValue` is exposed, but not `IntValue`, `LongValue`, etc.

Shane32:
Rather `ClrValue` and `TypedValue`, if there is a good reason for both, I would suggest this design pattern:

For `GraphQLValue`,

```cs
public object? ClrValue => GetClrValue();
protected abstract object? GetClrValue();
```

For value nodes such as `GraphQLStringValue`:

```cs
protected override object? GetClrValue() => ClrValue;
public new string ClrValue => _value ??= TypedValue;
```

Now `GraphQLValue.ClrValue` will return an `object?` while `GraphQLStringValue.ClrValue` will return a `string?`.

Not necessary but just an idea.

Shane32:
I guess an int scalar could return one of a few different types, such as `int` or `long` or `BigInteger`.  As such, it wouldn't make sense for it to have a `TypedValue` property.  Ok that makes sense.

Shane32:
I might recommend that we limit the amount of types that `GraphQLIntValue` can return.  For example, only `int`, `long`, `ulong` and `BigInteger`.  All other constructors should be removed.  Then it will notably simplify parsing in `IntGraphType`.

src/GraphQLParser/AST/GraphQLName.cs
sungam3r: Not actual anymore.

src/GraphQLParser/AST/Interfaces/IHasValueNode.cs
```diff
@@ -0,0 +1,21 @@
...
+/// <item><see cref="GraphQLFloatValue"/></item>
+/// <item><see cref="GraphQLStringValue"/></item>
+/// </list>
+/// </summary>
+public interface IHasValueNode
+{
+    /// <summary>
+    /// Value of AST node represented as <see cref="ROM"/>.
+    /// </summary>
+    public ROM Value { get; set; }
```
sungam3r, line 20: 😮 ! @Shane32  I accidentally wrote `public` and it works!

sungam3r: https://jeremybytes.blogspot.com/2019/11/c-8-interfaces-public-private-and.html

src/GraphQLParser/AST/Values/GraphQLNullValue.cs
```diff
@@ -6,11 +6,19 @@ namespace GraphQLParser.AST;
...
 {
     /// <inheritdoc/>
     public override ASTNodeKind Kind => ASTNodeKind.NullValue;
 
+    /// <summary>
+    /// Creates a new instance of <see cref="GraphQLNullValue"/>.
+    /// </summary>
+    public GraphQLNullValue()
+    {
+        Value = "null";
```
sungam3r, line 19: Default ROM that may be overridden by parser code. Added for convinience.

src/GraphQLParser.Tests/Values/GraphQLFloatValueTests.cs
```diff
@@ -0,0 +1,47 @@
...
+        var ex = Should.Throw<ArgumentOutOfRangeException>(() => new GraphQLFloatValue(double.NaN));
+        ex.Message.ShouldStartWith("Value cannot be NaN.");
+    }
+
+    [Fact]
+    public void FloatValue()
+    {
+        var value = new GraphQLFloatValue(1.1f);
+        value.Value.Length.ShouldBe(16);
+        value.Value.ShouldBe("1.10000002384186");
```
sungam3r, line 29: @Shane32 What do you think ?

src/GraphQLParser.Tests/Values/GraphQLIntValueTests.cs
```diff
@@ -0,0 +1,104 @@
...
+        var value = new GraphQLIntValue(1234554321m);
+        value.Value.Length.ShouldBe(10);
+        value.Value.ShouldBe("1234554321");
+    }
+
+    [Fact]
+    public void DecimalBad()
+    {
+        var ex = Should.Throw<ArgumentOutOfRangeException>(() => new GraphQLIntValue(12.3m));
+        ex.Message.Replace(",", ".").ShouldStartWith("Invalid integer number 12.3");
```
sungam3r, line 102: > Replace(",", ".")

ru-RU culture locally

src/GraphQLParser/AST/Values/GraphQLValue.cs
sungam3r: Reverted.

src/GraphQLParser/Shim/Numbers.cs
sungam3r: Reverted.

src/GraphQLParser/Shim/DecimalData.cs
sungam3r: reverted

src/GraphQLParser/ParserContext.Parse.cs
sungam3r: reverted

src/GraphQLParser/AST/GraphQLIntValue.cs
sungam3r: Reverted

src/GraphQLParser/AST/Values/GraphQLValue.cs
sungam3r: reverted

src/GraphQLParser/ParserContext.Parse.cs
sungam3r: I added `IHasValueNode` interface for all "primitive" nodes.

src/GraphQLParser/AST/Values/GraphQLListValue.cs
sungam3r: 👍  Moved in main repo.

src/GraphQLParser/AST/Values/GraphQLListValue.cs
sungam3r: reverted

src/GraphQLParser/AST/Values/GraphQLObjectValue.cs
sungam3r: reverted

src/GraphQLParser.ApiTests/GraphQLParser.approved.txt
sungam3r: reverted

sungam3r:
> Merge if you like. Or I can review later tonight. But I think I understand generally what this PR is.

I'm glad I decided to wait. I reverted majority of my changes. Main repo works fine without `ClrValue`. Please review again.

sungam3r:
I'lll add a few tests for `GraphQLName` conversion a bit later to achive 100% diff coverage.

src/GraphQLParser.Tests/Values/GraphQLIntValueTests.cs
Shane32: I suggest we remove the decimal constructor.  The decimal data type is designed for holding floating point data.  If someone wants to create an integer value from a decimal number, it is their job to store it in a valid data type first; int, long, BigInteger, etc.

src/GraphQLParser.Tests/Values/GraphQLFloatValueTests.cs
```diff
@@ -0,0 +1,47 @@
...
+        var value = new GraphQLFloatValue();
+        value.Value.Length.ShouldBe(0);
+    }
+
+    [Fact]
+    public void NanValue()
+    {
+        var ex = Should.Throw<ArgumentOutOfRangeException>(() => new GraphQLFloatValue(double.NaN));
+        ex.Message.ShouldStartWith("Value cannot be NaN.");
+    }
```
Shane32, line 22: Duplicate for Infinity and NegativeInfinity

Shane32 (changes requested):
I completed a full review of this PR

src/GraphQLParser.Tests/Values/GraphQLIntValueTests.cs
sungam3r: How to deal with it here?
```c#
    public virtual GraphQLValue? ToAST(object? value)
        {
            var serialized = Serialize(value);
            return serialized switch
            {
                bool b => new GraphQLBooleanValue(b),
                byte b => new GraphQLIntValue(b),
                sbyte sb => new GraphQLIntValue(sb),
                short s => new GraphQLIntValue(s),
                ushort us => new GraphQLIntValue(us),
                int i => new GraphQLIntValue(i),
                uint ui => new GraphQLIntValue(ui),
                long l => new GraphQLIntValue(l),
                ulong ul => new GraphQLIntValue(ul),
                BigInteger bi => new GraphQLIntValue(bi),
                decimal d => Math.Truncate(d) == d ? new GraphQLIntValue(d) : new GraphQLFloatValue(d), <--------
                float f => new GraphQLFloatValue(f),
                double d => new GraphQLFloatValue(d),
                string s => new GraphQLStringValue(s),
                null => new GraphQLNullValue(),
                _ => throw new NotImplementedException($"Please override the '{nameof(ToAST)}' method of the '{GetType().Name}' scalar to support this operation.")
            };
        }
```

src/GraphQLParser/AST/Values/GraphQLIntValue.cs
```diff
@@ -0,0 +1,150 @@
...
+
+    /// <summary>
+    /// Creates a new instance with the specified value.
+    /// </summary>
+    public GraphQLIntValue(decimal value)
+    {
+        if (Math.Truncate(value) != value)
+            throw new ArgumentOutOfRangeException(nameof(value), $"Invalid integer number {value}");
+        Value = value.ToString(CultureInfo.InvariantCulture);
+    }
```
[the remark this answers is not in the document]
sungam3r: See ` public virtual GraphQLValue? ScalarGraphType.ToAST(object? value)`. What to do here?

src/GraphQLParser/AST/Values/GraphQLBooleanValue.cs
```diff
@@ -6,11 +6,26 @@ namespace GraphQLParser.AST;
...
+    public GraphQLBooleanValue()
+    {
+    }
+
+    /// <summary>
+    /// Creates a new instance with the specified value.
+    /// </summary>
+    public GraphQLBooleanValue(bool value)
+    {
+        Value = value ? "true" : "false";
```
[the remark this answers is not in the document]
sungam3r: 🤔

src/GraphQLParser.Tests/Values/GraphQLIntValueTests.cs
Shane32: Simple:

`decimal d => new GraphQLFloatValue(d),`

There is nothing incorrect about this.

src/GraphQLParser/AST/Values/GraphQLBooleanValue.cs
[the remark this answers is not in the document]
sungam3r:
```c#
public ROM Value
{
    get => "null";
    set => { }
}
```

for NullValue it work since 'null' is constant.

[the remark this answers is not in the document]
Shane32: Then leave it alone.

src/GraphQLParser/AST/Values/GraphQLIntValue.cs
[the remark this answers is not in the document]
Shane32: It should create a `GraphQLFloatValue`.  A decimal is a floating-point number.

src/GraphQLParser/AST/Values/GraphQLFloatValue.cs
```diff
@@ -0,0 +1,109 @@
...
+
+    private static double ValidateValue(double value)
+    {
+        // TODO: see https://github.com/graphql-dotnet/graphql-dotnet/pull/2379#issuecomment-800828568 and https://github.com/graphql-dotnet/graphql-dotnet/pull/2379#issuecomment-800906086
+        // if (double.IsNaN(value) || double.IsInfinity(value))
+        if (double.IsNaN(value))
+            throw new ArgumentOutOfRangeException(nameof(value), "Value cannot be NaN."); // Value cannot be NaN or Infinity.
+
+        return value;
+    }
```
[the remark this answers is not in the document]
Shane32: See https://spec.graphql.org/draft/#sec-Float

> Non-finite floating-point internal values (NaN and Infinity) cannot be coerced to Float and must raise a field error.

Similar text about input coercion.  It's pretty explicit that NaN and Infinity are not supported by GraphQL.

src/GraphQLParser/AST/Values/GraphQLIntValue.cs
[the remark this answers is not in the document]
sungam3r: OK.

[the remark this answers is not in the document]
sungam3r: done

src/GraphQLParser/AST/Values/GraphQLIntValue.cs
```diff
@@ -0,0 +1,150 @@
...
+
+    /// <summary>
+    /// Creates a new instance with the specified value.
+    /// </summary>
+    public GraphQLIntValue(byte value)
+    {
+        Value = value.ToString(CultureInfo.InvariantCulture);
+    }
```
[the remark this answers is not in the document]
sungam3r: done

src/GraphQLParser/AST/Values/GraphQLNullValue.cs
sungam3r: I removed setter for `ROM Value`.

src/GraphQLParser.Tests/Values/GraphQLFloatValueTests.cs
sungam3r: Thanks.

src/GraphQLParser/AST/Values/GraphQLFloatValue.cs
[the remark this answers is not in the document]
sungam3r: done

src/GraphQLParser.Tests/Values/GraphQLFloatValueTests.cs
sungam3r: done 30 seconds ago :)

src/GraphQLParser/AST/Values/GraphQLBooleanValue.cs
```diff
@@ -6,15 +6,25 @@ namespace GraphQLParser.AST;
 /// AST node for <see cref="ASTNodeKind.BooleanValue"/>.
 /// </summary>
 [DebuggerDisplay("GraphQLBooleanValue: {Value}")]
-public class GraphQLBooleanValue : GraphQLValue
+public class GraphQLBooleanValue : GraphQLValue, IHasValueNode
```
Shane32, line 9: Now that we do not allow changing a boolean, we could have `GraphQLBooleanValue` abstract, and have separate derived classes for true and false variations.  I don't really care either way but it could possibly save some tiny bit of memory.

Perhaps separate PR even if we do want to do that.

src/GraphQLParser/AST/Values/GraphQLBooleanValue.cs
```diff
@@ -6,15 +6,25 @@ namespace GraphQLParser.AST;
...
+    public GraphQLBooleanValue(bool value)
+    {
+        _value = value;
+    }
+
     /// <summary>
     /// Boolean value represented as <see cref="ROM"/>.
     /// </summary>
-    public ROM Value { get; set; }
+    public ROM Value => _value ? "true" : "false";
```
Shane32, line 27: I think we should add a property to return the underlying boolean.  Saves another comparison in the main project.

Shane32 (approved):
I'm okay either way on adding the clr property for BooleanValue.  All files reviewed.  (well, non-test files)

src/GraphQLParser/AST/Values/GraphQLBooleanValue.cs
sungam3r: I did not do this because there will be combinations with descendant internal classes . In any case, if desired, this can be done later.

sungam3r: #195

src/GraphQLParser/AST/Values/GraphQLBooleanValue.cs
sungam3r: done

Shane32 (approved):

Issue #173, opened 2022-01-08
Add advanced SDLWriter tests for descriptions and comments


sungam3r:
See https://app.codecov.io/gh/graphql-dotnet/parser/blob/master/src/GraphQLParser/Visitors/SDLWriter.cs

sungam3r:
#210 : comments printing simplified

sungam3r:
`SDLPrinter` 100% covered.

PR #200, opened 2022-01-18
Pretty print + fix comments

### LINKED ISSUE #155: Do something with newlines in SDLWriter
Unify newlines or add setting.

sungam3r:
Fixes #155 and partially #173 . SDLWriter began to work **much** better, I checked it on a variety of new tests. Nevertheless, I am absolutely sure that I still need to make changes and check all the situations (comments/descriptions on all node types, indentation, formatting).

### DIFF
--- a/src/GraphQLParser/NotNullWhenAttribute.cs
+++ b/src/GraphQLParser/NotNullWhenAttribute.cs
@@ -1,0 +1,23 @@
+#if NETSTANDARD2_0
+
+// https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/NullableAttributes.cs#L69
+namespace System.Diagnostics.CodeAnalysis;
+
+/// <summary>Specifies that when a method returns <see cref="ReturnValue"/>, the parameter will not be null even if the corresponding type allows it.</summary>
+[AttributeUsage(AttributeTargets.Parameter, Inherited = false)]
+internal sealed class NotNullWhenAttribute : Attribute
+{
+    /// <summary>Initializes the attribute with the specified return value condition.</summary>
+    /// <param name="returnValue">
+    /// The return value condition. If the method returns this value, the associated parameter will not be null.
+    /// </param>
+    public NotNullWhenAttribute(bool returnValue)
+    {
+        ReturnValue = returnValue;
+    }
+
+    /// <summary>Gets the return value condition.</summary>
+    public bool ReturnValue { get; }
+}
+
+#endif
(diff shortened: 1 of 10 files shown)


src/GraphQLParser/NotNullWhenAttribute.cs
```diff
@@ -0,0 +1,23 @@
+#if NETSTANDARD2_0
+
+// https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/NullableAttributes.cs#L69
+namespace System.Diagnostics.CodeAnalysis;
+
+/// <summary>Specifies that when a method returns <see cref="ReturnValue"/>, the parameter will not be null even if the corresponding type allows it.</summary>
+[AttributeUsage(AttributeTargets.Parameter, Inherited = false)]
+internal sealed class NotNullWhenAttribute : Attribute
```
sungam3r, line 8: Needed for `TryPeek`.

src/GraphQLParser/ParserContext.Parse.cs
```diff
@@ -524,31 +524,31 @@ private GraphQLField ParseField()
         // start of alias (if exists) equals start of field
         int start = _currentToken.Start;
 
-        var nameOrAliasComment = GetComment();
+        var comment = GetComment(); // Greedy parsing for comment here - we read comment for field itself, not for alias.
```
sungam3r, line 527: @Shane32 I changed commmentss parsing for `Field`. Now it corresponds to the rest of the code when the comment refers to the object that is parsing by this method. I also changed tests.

src/GraphQLParser/Visitors/SDLWriter.cs
```diff
@@ -976,3 +1056,25 @@ private static async ValueTask WriteEncodedString(TContext context, ROM value)
...
+
+/// <summary>
+/// Options for <see cref="SDLWriter{TContext}"/>.
+/// </summary>
+public class SDLWriterOptions
+{
+    /// <summary>
+    /// Write comments into the output.
+    /// </summary>
+    public bool WriteComments { get; set; }
```
Shane32, line 1068: Does 'false' match default behavior before this option existed?  If not, this would be a breaking change.

Shane32: Note: I'm happy with `false` as default for v5, but if it used to print comments before, it should be `true` v4 and have a note in the migration doc for v5.  I see this targets `master`.

Shane32: Wrong repo; was thinking of the ASTPrinter in graphql.net .  nevermind!

Shane32 (approved):

src/GraphQLParser/Visitors/SDLWriter.cs
sungam3r: Oops , I forgot about setting `true`. When I wrote tests I used default value :
изображение

sungam3r: > but if it used to print comments before

There was no "before". All this code has never been released. Let's keep `false` for default.

PR #202, opened 2022-01-21
Optimize location conversion

sungam3r:
I spent on writing and debugging this algorithm a bunch of time. Although this code is performed only when making exception messages, I wanted to get rid of the regular expression. No memory allocations at all now. If you check code before changes on a rather big query, then you will make sure that the work with a regular expression here generates an abnormal amount of memory allocations.

### DIFF
--- a/src/GraphQLParser/ROM.cs
+++ b/src/GraphQLParser/ROM.cs
@@ -4,7 +4,11 @@ namespace GraphQLParser;
 
 /// <summary>
 /// A wrapper around ReadOnlyMemory{char} allowing you to use simple syntax when working with it.
-/// <br/>
+/// <br/><br/>
+/// This struct is designed to unify work with objects of <see cref="ReadOnlyMemory{T}"/>,
+/// <see cref="ReadOnlySpan{T}"/> ans <see cref="string"/>. For example, you can directly
+/// pass the string object to all methods accepting <see cref="ROM"/>.
+/// <br/><br/>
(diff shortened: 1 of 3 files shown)


src/GraphQLParser/ROM.cs
```diff
@@ -4,7 +4,11 @@ namespace GraphQLParser;
 
 /// <summary>
 /// A wrapper around ReadOnlyMemory{char} allowing you to use simple syntax when working with it.
-/// <br/>
+/// <br/><br/>
+/// This struct is designed to unify work with objects of <see cref="ReadOnlyMemory{T}"/>,
+/// <see cref="ReadOnlySpan{T}"/> ans <see cref="string"/>. For example, you can directly
```
sungam3r, line 9: @Shane32 I added these notes to make ROM usage by other projects more understandable.

sungam3r:
Example:
```
        string d = "KitchenSink".ReadGraphQLFile();

        for (int i = 0; i < 100; ++i)
        {
            var m1 = GC.GetAllocatedBytesForCurrentThread();
            var loc = new Location(d, 3000);
            var m2 = GC.GetAllocatedBytesForCurrentThread();
            var delta = m2 - m1;
        }
```

`var matches = _lineRegex.Matches((string)source);` returns collection with 152 items. Each loop iteration increases memory  by 35968 bytes!

sungam3r:
I'll add tests for 100% coverage later, 3:30AM here.

Shane32 (approved):

sungam3r:
The desire to get full coverage again made it possible to find a bug.

PR #203, opened 2022-01-22
Use string.GetHashCode for NET6

sungam3r:


Shane32:
You changed the ROM implementation, but your demo test does not use the ROM.GetHashCode() method.  It is using `ReadOnlyMemory<char>.GetHashCode`.  The test passes just fine with this change to the test:

```cs
        var h1 = new ROM(s1).GetHashCode();
        var h2 = new ROM(s2).GetHashCode();
```

Shane32:
The test also passes like this:
```cs
        var h1 = string.GetHashCode(s1.Span);
        var h2 = string.GetHashCode(s2.Span);
```

src/GraphQLParser/GraphQLParser.csproj
```diff
@@ -4,7 +4,7 @@
     <Description>Library containing lexer and parser for GraphQL syntax</Description>
     <Product>GraphQL Parser</Product>
     <AssemblyTitle>GraphQL Parser for .NET</AssemblyTitle>
-    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;net6</TargetFrameworks>
```
sungam3r, line 7: The package will increase by a third.

sungam3r:
Yes my mistake.

Shane32 (approved):

PR #210, opened 2022-01-23
Stop merging comments into one

sungam3r:
fixes https://github.com/graphql-dotnet/graphql-dotnet/issues/2878

Location of "merged" node was also wrong. It's correct (and more simple) not to merge comments into one node.


src/GraphQLParser.ApiTests/GraphQLParser.approved.txt
```diff
@@ -186,14 +186,12 @@ namespace GraphQLParser.AST
         public GraphQLDirectives() { }
         public override GraphQLParser.AST.ASTNodeKind Kind { get; }
     }
-    public class GraphQLDocument : GraphQLParser.AST.ASTNode, System.IDisposable
+    public class GraphQLDocument : GraphQLParser.AST.ASTNode
```
Shane32, line 189: No `IDisposable` 🥇

src/GraphQLParser.ApiTests/GraphQLParser.approved.txt
```diff
@@ -186,14 +186,12 @@ namespace GraphQLParser.AST
...
     {
         public GraphQLDocument() { }
         public System.Collections.Generic.List<GraphQLParser.AST.ASTNode> Definitions { get; set; }
         public override GraphQLParser.AST.ASTNodeKind Kind { get; }
-        public System.Collections.Generic.List<GraphQLParser.AST.GraphQLComment>? UnattachedComments { get; set; }
-        public void Dispose() { }
-        protected virtual void Dispose(bool disposing) { }
+        public System.Collections.Generic.List<System.Collections.Generic.List<GraphQLParser.AST.GraphQLComment>>? UnattachedComments { get; set; }
```
Shane32, line 194: Looks weird, but okay.  Was hoping that with the new AST classes perhaps this would disappear.

Shane32: Or that perhaps the unattached comments just get added to the list with the others...

Shane32 (approved):

sungam3r:
@Shane32 I need your help with encoding-decoding stuff for comments.

The code coverage dropped, see.
изображение

If every comment is now located in a separate AST node, then can we remove these branches?

src/GraphQLParser.ApiTests/GraphQLParser.approved.txt
sungam3r: Not a big deal. Let's keep unattached comments as is.

sungam3r:
I am always confused and forget what and how it should be decoded.

Shane32:
I can handle the decoding.

Yes, GraphQL's comments are per-line always.  There are no multi-line comments, as far as I know.  So if each line is a separate comment, then sure, we can just add them to a list rather than looking for line terminators and allocations and all that

Shane32:
I'll be happy to review all the decoding and be sure we have 100% coverage for that.

Shane32:
Comments have no encoding.  Just, starts at `#` and ends at line-terminator

sungam3r:
OK, then go on with this PR. 3 AM here, I'm done. Rel : #173

sungam3r:
Comments printing was simplified:
изображение

Single `ROM` is printed now.

sungam3r:
I removed implicit operator Memory -> ROM since Memory is always converted (implicitly) to ReadOnlyMemory first and then to ROM.

PR #215, opened 2022-01-25
Respect CancellationToken by ASTVisitor

sungam3r:
I just forgot 🤦‍♂️ .


src/GraphQLParser.Tests/Visitors/ASTVisitorTests.cs
```diff
@@ -0,0 +1,60 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using GraphQLParser.AST;
+using GraphQLParser.Visitors;
+using Shouldly;
+using Xunit;
+
+namespace GraphQLParser.Tests.Visitors;
```
sungam3r, line 9: It's renamed `DefaultNodeVisitorTests.cs` , I only added `ASTVisitor_Should_Respect_CancellationToken`

src/GraphQLParser/Visitors/ASTVisitor.cs
```diff
@@ -598,6 +598,8 @@ protected virtual async ValueTask VisitInputObjectTypeExtensionAsync(GraphQLInpu
     /// <param name="context">Context passed into all INodeVisitor.VisitXXX methods.</param>
     public virtual ValueTask VisitAsync(ASTNode? node, TContext context)
     {
+        context.CancellationToken.ThrowIfCancellationRequested();
```
sungam3r, line 601: One-line fix :)

Shane32: Is this ever called recursively?

Shane32 (approved):
See other comments

src/GraphQLParser.Tests/Visitors/SDLWriterTests.cs
```diff
@@ -664,7 +664,7 @@ public async Task WriteDocumentVisitor_Should_Throw_On_Unknown_Values(string tex
 
     private class Context : IASTVisitorContext
     {
-        public CancellationToken CancellationToken => throw new NotImplementedException();
+        public CancellationToken CancellationToken { get; set; }
```
Shane32, line 667: :+1:

src/GraphQLParser/Visitors/ASTVisitor.cs
sungam3r: > Is this ever called recursively?

Of course.

Shane32: Then I’d probably not add it.

Shane32: Whatever you think. I’ve never timed how slow that method is.

sungam3r: ?

sungam3r: That method dispatches call to some other method and so on.

PR #216, opened 2022-01-26
Optimize WriteEncodedStringAsync

sungam3r:
Should be much faster in majority of cases.

### DIFF
--- a/src/GraphQLParser/Visitors/SDLPrinter.cs
+++ b/src/GraphQLParser/Visitors/SDLPrinter.cs
@@ -998,11 +998,16 @@ public class SDLPrinter<TContext> : ASTVisitor<TContext>
     {
         await context.WriteAsync("\"").ConfigureAwait(false);
 
+        int startIndexOfNotEncodedString = 0;
         for (int i = 0; i < value.Span.Length; ++i)
         {
             char code = value.Span[i];
             if (code < ' ')
             {
+                if (startIndexOfNotEncodedString != i)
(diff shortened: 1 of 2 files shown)


src/GraphQLParser/Visitors/SDLPrinter.cs
```diff
@@ -1017,13 +1022,25 @@ private static async ValueTask WriteEncodedStringAsync(TContext context, ROM val
...
+            }
             else if (code == '"')
+            {
+                if (startIndexOfNotEncodedString != i)
+                    await context.WriteAsync(value.Slice(startIndexOfNotEncodedString, i - startIndexOfNotEncodedString)).ConfigureAwait(false);
+                startIndexOfNotEncodedString = i + 1;
+
                 await context.WriteAsync("\\\"").ConfigureAwait(false);
-            else
-                await context.WriteAsync(value.Slice(i, 1)/*code*/).ConfigureAwait(false); // TODO: no method for char
```
sungam3r, line 1024: Hot path: await + one-symbol slicing 

@Shane32 We should use the same approach for comments/block strings

sungam3r: So now in case of "normal" strings this loop does nothing just incrementing counter - single write is done at the end of method:
`
 await context.WriteAsync(value.Slice(startIndexOfNotEncodedString, value.Span.Length - startIndexOfNotEncodedString)).ConfigureAwait(false);
`

PR #226, opened 2022-01-31
Optimize escaped unicode case

sungam3r:


src/GraphQLParser/Visitors/SDLPrinter.cs
```diff
@@ -1164,7 +1170,10 @@ private static async ValueTask WriteEncodedStringAsync(TContext context, ROM val
                 else if (code == '\t')
                     await context.WriteAsync("\\t").ConfigureAwait(false);
                 else
-                    await context.WriteAsync("\\u" + ((int)code).ToString("X4")).ConfigureAwait(false);
+                {
+                    await context.WriteAsync("\\u00").ConfigureAwait(false);
```
sungam3r, line 1174: We are under `code < ' '` condition and high byte is always `0`.

Shane32: Why not add "\\u00" to each string in the array so we eliminate a call to WriteAsync?

src/GraphQLParser/Visitors/SDLPrinter.cs
```diff
@@ -1138,6 +1138,12 @@ private static bool TryPeekParent(TContext context, [NotNullWhen(true)] out ASTN
         return true;
     }
 
+    private static readonly string[] _hex32 = new[]
+    {
+        "00", "01", "02", "03", "04", "05", "06", "07", "08", "09", "0A", "0B", "0C", "0D", "0E", "0F",
+        "10", "11", "12", "13", "14", "15", "16", "17", "18", "19", "1A", "1B", "1C", "1D", "1E", "1F"
```
Shane32, line 1144: If these are the equivalent of individual constants, shouldn't we test for all conditions?

sungam3r: done

src/GraphQLParser/Visitors/SDLPrinter.cs
sungam3r: done

Shane32 (approved):

PR #230, opened 2022-02-01
Add tests

sungam3r:
@Shane32  Please review this test for block string. Also I have a question about one case left:
изображение

Is such a situation possible? If so, could you give an example of a test that should be added? If no I suggest to remove that case.

### DIFF
--- a/src/GraphQLParser.Tests/Visitors/SDLPrinterTests.cs
+++ b/src/GraphQLParser.Tests/Visitors/SDLPrinterTests.cs
@@ -707,6 +707,14 @@ multilined
 "a \u0010\u0011\u0012\u0013\u0014\u0015\u0016\u0017\u0018\u0019\u001A\u001B\u001C\u001D\u001E\u001F b",
 @"""a \u0010\u0011\u0012\u0013\u0014\u0015\u0016\u0017\u0018\u0019\u001A\u001B\u001C\u001D\u001E\u001F b""
 ")]
+    [InlineData(                    // TODO: Change test condition?
+"Test\r\nLine 2\rLine 3\nLine 4",
+@"""""""
+Test
+Line 2Line 3
+Line 4
+""""""
+")]
     public async Task Description_With_Escaped_Unicode_Should_Be_Printed(string text, string expected)
     {
         var description = new GraphQLDescription(text);
--- a/src/GraphQLParser.Tests/Visitors/SDLPrinterTests.cs
+++ b/src/GraphQLParser.Tests/Visitors/SDLPrinterTests.cs
@@ -580,6 +580,8 @@ type Query {
     [InlineData(28, "Test \\n escaping", "Test \\\\n escaping", false)]
     [InlineData(29, "Test \\u1234 escaping", "Test \\\\u1234 escaping", false)]
     [InlineData(30, "Test \\ escaping", "Test \\\\ escaping", false)]
+    [InlineData(31, "t\"est\n  line2", "\nt\"est\nline2\n", true)] // BlockString with double quote inside
+    [InlineData(32, "t\\\"\"\"est\n  line2", "\nt\\\"\"\"est\nline2\n", true)] // BlockString with triple double quote inside
     public async Task SDLPrinter_Should_Print_BlockStrings(int number, string input, string expected, bool isBlockString)
     {
         number.ShouldBeGreaterThan(0);
--- a/src/GraphQLParser/Exceptions/GraphQLParserException.cs
+++ b/src/GraphQLParser/Exceptions/GraphQLParserException.cs
@@ -32,6 +32,7 @@ public class GraphQLParserException : Exception
         Location = location;
     }
 
+    //TODO: all this code can be rewritten to not allocate strings but is it worth it?
     private static string ComposeMessage(string description, ReadOnlySpan<char> source, Location location)
     {
         return $"Syntax Error GraphQL ({location.Line}:{location.Column}) {description}" +
@@ -46,7 +47,7 @@ public class GraphQLParserException : Exception
         string nextLineNum = (line + 1).ToString();
         int padLen = nextLineNum.Length;
         string[] lines = source
-            .ToString() // TODO: heap allocation
+            .ToString()
             .Split(new string[] { "\n" }, StringSplitOptions.None)
             .Select(e => ReplaceWithUnicodeRepresentation(e))
             .ToArray();
@@ -79,7 +80,7 @@ public class GraphQLParserException : Exception
         {
             if (IsReplacementCharacter(code))
             {
-                buffer.Append(GetUnicodeRepresentation(code));
+                buffer.Append("\\u").Append(((int)code).ToString("D4"));
             }
             else
             {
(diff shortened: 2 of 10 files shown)


sungam3r:
Only one line uncovered
изображение

Shane32:
I’ll try to review tomorrow or maybe tonight.

sungam3r:
Also found this
изображение

I repeat, I never had a patience to understand all these transcoding 😞 .

Shane32:
I really enjoy those types of coding puzzles. But I’ve been concentrating on the auto registering graph types

src/GraphQLParser/Exceptions/GraphQLParserException.cs
```diff
@@ -102,10 +103,4 @@ private static bool HasReplacementCharacter(string str)
     }
 
     private static bool IsReplacementCharacter(char code) => code < 0x0020 && code != 0x0009 && code != 0x000A && code != 0x000D;
-
-    private static string GetUnicodeRepresentation(char code) => code switch
-    {
-        '\0' => "\\u0000",
```
sungam3r, line 108: I don't know why does this line here. `((int)'\0').ToString("D4")` gives `"\\u0000"` as well.

src/GraphQLParser/ParserContext.Parse.cs
```diff
@@ -1328,22 +1328,20 @@ private GraphQLType ParseType()
...
+
+            nonNull.Type = type;
+            // move comment from wrapped type to wrapping type
+            nonNull.Comments = type.Comments;
+            type.Comments = null;
+            nonNull.Location = GetLocation(start);
+            type = nonNull;
         }
 
-        var nonNull = NodeHelper.CreateGraphQLNonNullType(_ignoreOptions); //TODO: deal with depth
```
sungam3r, line 1337: TODO resolved, nothing to fix here, I just invert condition to align count of `IncreaseDepth` calls with count of `DecreaseDepth` calls.

Shane32:
I spent some time looking at this.  I think what we need to decide is whether or not we want strings to always parse to exactly as input.

For instance, there is no way to encode any of the following text in a blockstring:
- `"\nTest"` - first line contains only whitespace
- "Test\n" - last line contains only whitespace
- `"Line 1\r\nLine 2"` - newlines (meaning `"\r\n", "\r", and "\n") coalesce to `\n` within a blockstring
- Strings containing control characters except certain ones

It is also noteworthy that some string sequences require a trailing blank line when formatted as a blockstring.  Specifically:
- Any string that ends with a quote must be encoded with a trailing newline in order to be properly parsed.
- If the text has uniform indentation

If we assume that the initial `"""` and trailing `"""` always is on its own line, with the trailing `"""` indented properly, then we can simplify checks a bit.  The rules now become something like this (to ensure that a string will parse exactly as input):
- Must not have beginning or trailing lines that are empty or contain only whitespace (tab or space)
- Cannot contain `"\r"` at all
- Cannot contain control characters unsupported by the GraphQL spec as SourceCharacter

I suggest we add an option to the printer to allow for some or all of the following:
- Allow newline conversion (`"\r\n"` and `"\r"` ->`"\n"`)
- Allow removing starting or ending blank lines

I'm not sure exactly what to call these options, and if we should have them separate or a single option.  Also I would assume that when we print an AST, only the description would be subject to such changes.  I would expect that printed string literals as part of a query, for instance, would always ensure that they can parse to exactly what was input.

Shane32:
As for the line of code in question, the input should be `"Test\r\nLine 2\rLine 3\nLine 4"` and the output should be:

```
"""
Test
Line 2
Line 3
Line 4
"""
```

Unless I miss my guess, however, this would print something like this currently:

```
"""
Test
Line 2Line 3
Line 4
"""
```

It would "cover" the line of code you mentioned, but I would consider that behavior incorrect -- but there is no way to encode `"\r"` in a blockstring at all, as previously discussed.  So even if this is "fixed", the output still will not match the input, and would produce `"Test\nLine 2\nLine 3\nLine 4"` when decoded.

src/GraphQLParser/Exceptions/GraphQLParserException.cs
Shane32: Probably just an optimization for null characters to prevent an allocation or something.

src/GraphQLParser.Tests/Visitors/SDLPrinterTests.cs
```diff
@@ -580,6 +580,8 @@ type Query
     [InlineData(28, "Test \\n escaping", "Test \\\\n escaping", false)]
     [InlineData(29, "Test \\u1234 escaping", "Test \\\\u1234 escaping", false)]
     [InlineData(30, "Test \\ escaping", "Test \\\\ escaping", false)]
+    [InlineData(31, "t\"est\n  line2", "\nt\"est\nline2\n", true)] // BlockString with double quote inside
+    [InlineData(32, "t\\\"\"\"est\n  line2", "\nt\\\"\"\"est\nline2\n", true)] // BlockString with triple double quote inside
     public async Task SDLPrinter_Should_Print_BlockStrings(int number, string input, string expected, bool isBlockString)
```
Shane32, line 585: These new test conditions look good.

We cannot test for a string with a CR in it because all strings here are interpreted as a blockstring.  Since a blockstring cannot contain a CR (it would be parsed as a LF), it is impossible to test here what happens when the code tries to serialize a string containing a CR.  We would need to add it to another test.  We also cannot test for beginning or trailing whitespace and so on.  All that we can test for here is strings that can be encoded as blockstrings to begin with.

src/GraphQLParser/Visitors/PrintContextExtensions.cs
```diff
@@ -21,7 +21,7 @@ public static ValueTask WriteAsync<TContext>(this TContext context, ROM value)
         var task =
 #if NETSTANDARD2_0
         // no cancellationToken support on netstandard2.0
-        context.Writer.WriteAsync(value.ToString());
+        context.Writer.WriteAsync(value.ToString()); //ISSUE: allocation - either WriteAsync(value.ToString()) or Write(char value) in a loop
```
Shane32, line 24: Calling `Write(char value)` in a loop would be a performance killer.  Much better to have an allocation if needed.

Shane32 (approved):

src/GraphQLParser/Visitors/PrintContextExtensions.cs
sungam3r: ok

sungam3r:
@Shane32  Thanks for test example. I added it to get full coverare (correct or not). Unfortunately, I understood little from your explanation. I confess that the transcoding is not my strength. Let's add the necessary options and change the output if you consider it necessary.

src/GraphQLParser/Exceptions/GraphQLParserException.cs
sungam3r: Looks strange among all other string allocations above.

sungam3r:
`SDLPrinter` 100% covered

PR #236, opened 2022-02-07
Fix docs

sungam3r:
Also tests https://github.com/github/codeql/issues/7544#issuecomment-1031161200
before:
изображение


sungam3r:
https://github.com/github/codeql/issues/7544#issuecomment-1031288773

PR #240, opened 2022-02-10
Test blockstring linefeed parsing

Shane32:


Shane32:
@sungam3r Should be covered now.

PR #259, opened 2022-07-04
Define ParserContext as ref struct

Shane32:


src/GraphQLParser/ParserContext.cs
```diff
@@ -6,7 +6,7 @@
 namespace GraphQLParser;
 
 // WARNING: mutable struct, pass it by reference to those methods that will change it
-internal partial struct ParserContext
+internal ref partial struct ParserContext
```
sungam3r, line 9:
```suggestion
internal partial ref struct ParserContext

```

src/GraphQLParser/ParserContext.Parse.cs
```diff
@@ -4,7 +4,7 @@
 namespace GraphQLParser;
 
 // WARNING: mutable struct, pass it by reference to those methods that will change it
-internal partial struct ParserContext
+internal ref partial struct ParserContext
```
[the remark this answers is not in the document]
sungam3r: oops, CS1585	Member modifier 'ref' must precede the member type and name

src/GraphQLParser/ParserContext.Parse.cs
```diff
@@ -4,7 +4,7 @@
 namespace GraphQLParser;
 
 // WARNING: mutable struct, pass it by reference to those methods that will change it
```
sungam3r, line 6:
```suggestion
// WARNING: mutable ref struct, pass it by reference to those methods that will change it
```

src/GraphQLParser/ParserContext.cs
```diff
@@ -6,7 +6,7 @@
 namespace GraphQLParser;
 
 // WARNING: mutable struct, pass it by reference to those methods that will change it
```
sungam3r, line 8:
```suggestion
// WARNING: mutable ref struct, pass it by reference to those methods that will change it

```

sungam3r (approved):
🤔 It seemed to me that I once tried to do this. Do you have a practical scenario when it is required? Let's add ref to `LexerContext` too.

Shane32:
> 🤔 It seemed to me that I once tried to do this. Do you have a practical scenario when it is required? Let's add ref to `LexerContext` too.

No. It simply is already a ref struct — it is always (and must be) passed by ref; it’s a struct, and it’s internal so the change does not affect the public api or operation but is simply “good housekeeping”

I didn’t look at LexerContext

Shane32:
The change to ref struct prevents accidental storage on the heap, such as seen by the change to the tests

PR #262, opened 2022-08-03
Feature/assembly strongname

anderslaub:
## Motivation

Many solutions based on .NET Framework (net4x) has both direct and inherited dependencies on GraphQl-dotnet.  

Since the currently released assemblies does not have a strongly typed name, developers who work on such solutions are forced to depend on a compatible version of graphql-dotnet as their inherited dependency. This often means v2.4.0 due to breaking changes introduced in later major versions.

By strong name signing the assemblies that target netstandard2.0/1, developers who are currently stuck on an old inherited version will be able to do codebase versioned assembly binding.

_example:_
```xml
            <dependentAssembly xdt:Transform="InsertIfMissing"  xdt:Locator="Condition(asmv1:assemblyIdentity/@name='GraphQl')">
                <assemblyIdentity name="GraphQl"  publicKeyToken="e6238258560628ee" />
                <codeBase version="5.0.0.0" href="bin\deps\GraphQL.dll" />
            </dependentAssembly>

            <dependentAssembly xdt:Transform="InsertIfMissing"  xdt:Locator="Condition(asmv1:assemblyIdentity/@name='GraphQl.DataLoader')">
                <assemblyIdentity name="GraphQl.DataLoader" publicKeyToken="e6238258560628ee" />
                <codeBase version="5.0.0.0" href="bin\deps\GraphQl.DataLoader.dll" />
            </dependentAssembly>

            <dependentAssembly xdt:Transform="InsertIfMissing"  xdt:Locator="Condition(asmv1:assemblyIdentity/@name='GraphQl.NewtonsoftJson')">
                <assemblyIdentity name="GraphQL.NewtonsoftJson" publicKeyToken="e6238258560628ee" />
                <codeBase version="5.0.0.0" href="bin\deps\GraphQL.NewtonsoftJson.dll" />
            </dependentAssembly>

            <dependentAssembly xdt:Transform="InsertIfMissing"  xdt:Locator="Condition(asmv1:assemblyIdentity/@name='GraphQLParser')">
                <assemblyIdentity name="GraphQLParser" publicKeyToken="e6238258560628ee" />
                <codeBase version="8.0.0.0" href="bin\deps\GraphQLParser.dll" />
            </dependentAssembly>
```

Signing the assemblies also enable other use-cases which has otherwise been prevented by legacy requirements for strong-named assembly (such as Visual Studio extension development).  

The lack of a strong name on current releases prevents developers from transitioning their codebases to later versions of GraphQl leaving many projects stuck on graphql-dotnet v2.4.0

As support for this thesis, [the nuget.org download stats](https://www.nuget.org/packages/GraphQL/2.4.0) clearly shows that v2.4.0 is by far the most downloaded nuget and continues to have a very significant amount of daily downloads.

## Notes
Pushing the private key used for signing the assemblies is the recommended approach for open-source projects and is not a [security concern](https://docs.microsoft.com/en-us/dotnet/standard/assembly/strong-named).  

We can all agree that signing assemblies for managing dependencies should never have been a thing or should at least be a thing of the past. This is unfortunately not the reality yet and leaving this out on popular frameworks such as this makes modernizing code-bases even harder.

## Considerations
Assembly sigining could alternatively be done in a Github action.  

The reason for having it as part of the build is that it provide more flexibility and also enable local builds to be signed with the same key as the officially released assemblies - making testing and extending less prone to error.  

## Important
Since strong named assemblies only can reference other strongly named assemblies. These 2 PRs are dependent:
- https://github.com/graphql-dotnet/parser/pull/262
- https://github.com/graphql-dotnet/graphql-dotnet/pull/3265

## References

Microsoft recommend publicly published .NET libraries and open-source maintainers to sign their assemblies for best cross-runtime compatibility.
- https://github.com/dotnet/runtime/blob/main/docs/project/strong-name-signing.md
- https://github.com/dotnet/project-system/blob/main/eng/imports/StrongName.targets


sungam3r:
@anderslaub I need some time to tjing about this PR. I saw similar discussions in other projects. Opinions were different.

anderslaub:
> @anderslaub I need some time to tjing about this PR. I saw similar issues/discussions in other projects. Opinions were different.

@sungam3r sure of course, and thanks for looking into this. 

I am personally not aware of any reasons not to sign assemblies in a framework like this - especially when they are still being heavily used by solutions targeting  .NET Framework such as this. 

It might be that there are some details that I am not aware of? Otherwise it just means forcing solutions to depend on old versions dictated by their dependencies thus block adoption of new versions by a still very large part of the .net community.

Shane32:
I looked through much of the available MS documentation on this topic.  I cannot find any significant reason not to sign the assembly.

Here's what I noticed:
* On .NET Framework, having the assembly strongly named is useful to other strongly-named assemblies
* On .NET Framework, if we sign the assembly, it will require binding redirects in some situations (probably often), which are generally automatically created by Visual Studio
* All .NET Core assemblies are strongly named
* Having the private key stored in the repository is recommended and desired
* It should not affect .NET Core applications, per the snip below:

> For .NET Core and .NET 5+, strong-named assemblies do not provide material benefits. The runtime never validates the strong-name signature, nor does it use the strong-name for assembly binding.

I support these changes and would suggest that we release it as 8.1.0 prior to GraphQL v7 release.

src/_SignAssembly.targets
```diff
@@ -0,0 +1,40 @@
...
+  <!--
+    NOTE: GraphQLParser.Tests depends on netstandard 2.1 for netcore3.1 and net5.0 tests, requiring a signed friend assembly for internals visible to.
+  -->
+  <PropertyGroup Condition="'$(TargetFramework)'=='net6'">
+    <SignAssembly>false</SignAssembly>
+  </PropertyGroup>
+
+  <PropertyGroup  Condition="'$(TargetFramework)'!='net6'">
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>$(SolutionDir)graphql.snk</AssemblyOriginatorKeyFile>
```
Shane32, line 17:
```suggestion
    <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)graphql.snk</AssemblyOriginatorKeyFile>
```
Maybe this will help the codeql to build properly

src/_SignAssembly.targets
```diff
@@ -0,0 +1,40 @@
...
+      
+        Ref: https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/strong-name-signing-exceptions#recommended-action
+     -->
+
+  <!--
+    NOTE: GraphQLParser.Tests depends on netstandard 2.1 for netcore3.1 and net5.0 tests, requiring a signed friend assembly for internals visible to.
+  -->
+  <PropertyGroup Condition="'$(TargetFramework)'=='net6'">
+    <SignAssembly>false</SignAssembly>
+  </PropertyGroup>
```
Shane32, line 13: MS assemblies compiled for .NET 6 are still signed.  It's just that .NET Core ignores them.  I would do the same here.

image

image

Shane32:
@anderslaub If you're available to help finish up a few remaining comments on this PR, I'll get it merged in.  Let's do it asap as I would like to include this in the 7.0 release of GraphQL.NET.

src/_SignAssembly.targets
anderslaub: I know and I considered always signing too. But then I decided to skip the net6.0 specific build since it was so easy to do.

According to MS then net6.0 will completely ignore if the assembly is signed, making it a waste of space, literally.

I can remove the conditional signing if you prefer?

Shane32: Let’s remove the conditions. It will be a simpler props file and matches what Microsoft does for their packages

Shane32: @anderslaub lmk when done and I can merge

anderslaub:
@Shane32 Done - sorry for the wait!

src/_SignAssembly.targets
anderslaub: @Shane32 Done. I've also moved the SignAssembly prop to directory.build.props so it gets set before loading the project props during build - just so it can be overridden in a project file if this will be needed at some later point for something (not sure why it would be).

Shane32 (approved):

Shane32:
@sungam3r I have not seen any comments from you, so I reviewed this PR and am merging it.  Will release as 8.1.0 with the GraphQL.NET 7.0 release.

Shane32:
I have verified that the packaged nuget file contains strong named assemblies.

sungam3r:
Was busy. Once upon a time 😄 , in era of development on full .NET Framework, I dealt with strong names, assembly signing, binding redirects and all that stuff. It was a PITA.

> The lack of a strong name on current releases prevents developers from transitioning their codebases to later versions of GraphQl **leaving many projects stuck on graphql-dotnet v2.4.0**

@anderslaub I don't understand. v2.4.0 has no strong name either:
изображение
Why are you so focusing on version 2.4?

> As support for this thesis, [the nuget.org download stats](https://www.nuget.org/packages/GraphQL/2.4.0) clearly shows that v2.4.0 is by far the most downloaded nuget and continues to have a very significant amount of daily downloads.

What is so special in version 2.4 that is associated with strong names and this PR?

> I have verified that the packaged nuget file contains strong named assemblies.

I see.

src/GraphQLParser.sln
```diff
@@ -15,9 +15,12 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		..\.gitignore = ..\.gitignore
 		Directory.Build.props = Directory.Build.props
 		Directory.Build.targets = Directory.Build.targets
+		graphql.snk = graphql.snk
```
sungam3r, line 18: private.snk ?

Shane32:
@sungam3r Please read this detailed comment by @anderslaub that explains the reasoning behind adding a strong name to 2.4.0:

https://github.com/graphql-dotnet/graphql-dotnet/pull/3265#issuecomment-1207451349

I'm not sure I follow it entirely, but it sounds like:

1. his root issue is due to a third-party vendor supplying a .NET library with a dependency on 2.4.0, while @anderslaub is using a much newer version of GraphQL.NET.

2. he has implemented a workaround, based on adding a strong name to the current version of GraphQL.NET.  (Presumably this workaround is only viable within a .NET Framework environment.)

3. adding a strong name to 2.4.0 could make this issue easier to deal with, presuming of course that the third-party vendor bumps their dependency to 2.4.1 (the version with the strong name)

@anderslaub correct me if I'm wrong.

As such, it seems that the reasoning for a strong name is justified, as is patching older versions of GraphQL.NET (to a lesser degree).  I wouldn't mind doing so for 3.x, 4.x or 5.x, which would be easy enough to do, but we do not have the GitHub Action workflows in the 2.4.0 branch.

Let's continue discussion over at https://github.com/graphql-dotnet/graphql-dotnet/pull/3265

Shane32:
Actually, we could not do so for the old versions quite so easily -- we would need to release strong-named parser libraries first.

src/GraphQLParser/GraphQLParser.csproj
```diff
@@ -7,6 +7,7 @@
     <TargetFrameworks>netstandard2.0;netstandard2.1;net6</TargetFrameworks>
     <PackageId>GraphQL-Parser</PackageId>
     <PackageTags>GraphQL;json;api;parser</PackageTags>
+    <_FriendAssembliesPublicKey>PublicKey=0024000004800000940000000602000000240000525341310004000001000100352162dbf27be78fc45136884b8f324aa9f1dfc928c96c24704bf1df1a8779b2f26c760ed8321eca5b95ea6bd9bb60cd025b300f73bd1f4ae1ee6e281f85c527fa013ab5cb2c3fc7a1cbef7f9bf0c9014152e6a21f6e0ac6a371f8b45c6d7139c9119df9eeecf1cf59063545bb7c07437b1bc12be2c57d108d72d6c27176fbb8</_FriendAssembliesPublicKey>
```
sungam3r, line 10: Why underscore ?

src/Directory.Build.props
```diff
@@ -23,6 +23,14 @@
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
     <PackageReadmeFile Condition="'$(IsPackable)' == 'true'">README.md</PackageReadmeFile>
+
+    <!--
+      Strong name signing of netstandard assemblies for better cross-runtime compatibility
```
sungam3r, line 28: Assembly for net6 TFM is also signed now.

src/_SignAssembly.targets
```diff
@@ -0,0 +1,32 @@
+<Project>
```
sungam3r, line 1: Is there any good reason to use the underscore in the file name?

src/Directory.Build.props
```diff
@@ -23,6 +23,14 @@
...
+
+        Ref: https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/strong-name-signing-exceptions#recommended-action
+     -->
+    <SignAssembly>true</SignAssembly>
+    <CustomBeforeMicrosoftCommonTargets>$(MSBuildThisFileDirectory)_SignAssembly.targets</CustomBeforeMicrosoftCommonTargets>
```
sungam3r, line 33: MSBuild project system with all these magic hidden properties and targets overriding each other is really hard to learn 😕

src/_SignAssembly.targets
```diff
@@ -0,0 +1,32 @@
...
+      <_DelaySignMainAssembly>$(DelaySign)</_DelaySignMainAssembly>
+      <DelaySign Condition="'$(PublicSign)' == 'true'">true</DelaySign>
+    </PropertyGroup>
+  </Target>
+
+  <Target Name="CleanupAfterDelaySigningSatelliteAssemblies" AfterTargets="GenerateSatelliteAssemblies" Condition="'$(MSBuildRuntimeType)' != 'Core' and '$(SignAssembly)' == 'true'">
+    <PropertyGroup>
+      <DelaySign>$(_DelaySignMainAssembly)</DelaySign>
+    </PropertyGroup>
+  </Target>
```
sungam3r, line 31: @anderslaub Why to bother with this workaround if GraphQL-Parser has no satellite assemblies ?

sungam3r: And if so this file can be removed and two lines of config just moved into Directory.Build.props 
```
 <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)graphql.snk</AssemblyOriginatorKeyFile>
 <PublicKeyToken>e6238258560628ee</PublicKeyToken>
```

I'm even not sure if `PublicKeyToken` property is needed since `AssemblyOriginatorKeyFile` contains full public key. How is `PublicKeyToken` property used?

src/Directory.Build.props
Shane32: agreed

src/_SignAssembly.targets
Shane32: I thought I read this was to support `InternalsVisibleTo` but I could be mistaken.

src/GraphQLParser.sln
```diff
@@ -15,9 +15,12 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
...
 		..\LICENSE.md = ..\LICENSE.md
 		..\assets\logo.64x64.png = ..\assets\logo.64x64.png
+		public.snk = public.snk
```
sungam3r, line 21: I think this file is not nessessary at all.

src/GraphQLParser.sln
sungam3r: I thought some more on this. `graphql.snk` is fine since we may copy-paste this file across all org repos.

src/_SignAssembly.targets
sungam3r: #265

src/GraphQLParser/GraphQLParser.csproj
anderslaub: It's a common naming convention for "internally" used msbuild props - a prefix for props that is not intended to be overridden.

You will see this naming convention widely used in MS props / targets files but it is afaik not well-described anywhere. I am just used to following this.

src/Directory.Build.props
anderslaub: True, it is however only signing the netstandard assemblies that is done for better cross-runtime compatibility.

src/_SignAssembly.targets
anderslaub: Same reply as above. It's by "convention" - not a well-described one so - I'm good with naming it whatever match the current solution naming. 

I would usually place msbuild files such as this in a subfolder but I did not want to introduce this for only 1 file.

Since we are now also signing the net6 assembly always, the content of the file could really just go in Directory.Build.targets if this is preferred? 

It is right now being loaded via the `CustomBeforeMicrosoftCommonTargets` prop which makes it possible to evaluate conditions based on properties that gets set in the individual csproj files (so it is loaded after the targets defined in these, directory.build.targets is loaded before)

src/_SignAssembly.targets
anderslaub: > @anderslaub Why to bother with this workaround if GraphQL-Parser has no satellite assemblies ?

This is copy-pasted from the referenced link and you're right it is irrelevant to include. We are also not doing PublicSign but real signing.

src/GraphQLParser.sln
anderslaub: The file is just included as reference since it is this public key that is used for friend assembly references. 

It is created from the private snk `graphql.snk` using `sn.exe` 

Let me know if I should delete it.

src/_SignAssembly.targets
anderslaub: > I'm even not sure if `PublicKeyToken` property is needed since `AssemblyOriginatorKeyFile` contains full public key. How is `PublicKeyToken` property used?

It is not needed for signing the assemblies at all and also not for the friend assembly references (InternalsVisibleTo). 

The property is merely set so it _could_ be used if needed during the build for referencing with or without public key. 

Let me know if I should remove it.

src/GraphQLParser.sln
anderslaub: that was exactly my thinking when naming it. I've used the same snk for signing in https://github.com/graphql-dotnet/graphql-dotnet/pull/3265 - unsure if there is some smarter way to share it between repos without much effort. 

Assemblies could of course just be signed in gh action rather than during build but then making "custom" builds from source would become a real pain for everyone.

Shane32:
@anderslaub Can you review @sungam3r 's changes in #265 ?  I assume they are consistent with the comments that you two have been making in this PR.

src/_SignAssembly.targets
sungam3r: Already removed in #365. Let's keep only needed settings in csproj files.

src/GraphQLParser.sln
sungam3r: Deleted in #265

sungam3r: I have no much experience in cryptography, but I suspected that a public key could be obtained from a private key with some standard utility.

src/GraphQLParser/GraphQLParser.csproj
sungam3r: ok

src/_SignAssembly.targets
sungam3r: I looked into some other repos to check how signing is managed here. It looks like `SignAssembly` and `AssemblyOriginatorKeyFile` properties are quite enough to do the job.

PR #265, opened 2022-08-10
Simplify assembly signing

sungam3r:
I hope it works the same as before.


Shane32:
Should _SignAssembly.targets be deleted in this PR?

Shane32:
I see it removed from the sln but not deleted

src/GraphQLParser/GraphQLParser.csproj
```diff
@@ -7,7 +7,7 @@
     <TargetFrameworks>netstandard2.0;netstandard2.1;net6</TargetFrameworks>
     <PackageId>GraphQL-Parser</PackageId>
     <PackageTags>GraphQL;json;api;parser</PackageTags>
-    <_FriendAssembliesPublicKey>PublicKey=0024000004800000940000000602000000240000525341310004000001000100352162dbf27be78fc45136884b8f324aa9f1dfc928c96c24704bf1df1a8779b2f26c760ed8321eca5b95ea6bd9bb60cd025b300f73bd1f4ae1ee6e281f85c527fa013ab5cb2c3fc7a1cbef7f9bf0c9014152e6a21f6e0ac6a371f8b45c6d7139c9119df9eeecf1cf59063545bb7c07437b1bc12be2c57d108d72d6c27176fbb8</_FriendAssembliesPublicKey>
```
anderslaub, line 10: Prefixing the property with an underscore to indicate that it is only intended for "internal" usage / scope is to me a common well-understood convention. 

I think most devs who has dived a bit deeper into msbuild will recognize this as the normal way to name properties that should not be changed outside the scope (or file sometimes) of where it has been set. 

Then again, it's just a name and it makes little sense if others do not understand the intended meaning.

The only reference I can find to semi backup my perception is the msbuild repo https://github.com/dotnet/msbuild/search?q=filename%3A*.targets+%3C_ where underscore prefix are widely used in this way.

sungam3r:
> Should _SignAssembly.targets be deleted in this PR?

🤦‍♂️ forgot

Shane32 (approved):

src/GraphQLParser/AST/GraphQLName.cs
```diff
@@ -36,12 +36,9 @@ public string StringValue
     {
         get
         {
-            if (_string == null)
-            {
-                _string = Value.Length == 0
+            _string ??= Value.Length == 0
```
sungam3r, line 39: new versions of NET SDK is more and more smart

src/GraphQLParser/GraphQLParser.csproj
sungam3r: OK, I returned underscore back.

sungam3r:
The same result as expected
изображение

PR #280, opened 2022-11-10
NET7

sungam3r:


.github/workflows/codeql-analysis.yml
```diff
@@ -16,10 +16,10 @@ jobs:
     - name: Checkout source
       uses: actions/checkout@v3
 
-    - name: Setup .NET Core SDK
-      uses: actions/setup-dotnet@v3
+    - name: Setup .NET SDK
+      uses: actions/setup-dotnet@v3.0.3
```
Shane32, line 20: Why pinning to 3.0.3?

sungam3r: Why not?  Just updated as here https://github.com/graphql-dotnet/graphql-dotnet/blob/master/.github/workflows/codeql-analysis.yml#L20

Shane32: Because then you won’t get v3.0.4

sungam3r: I don't understand. Dependabot bumps this version from time to time: 2 -> 3 -> 3.0.1 -> 3.0.2 -> 3.0.3

Shane32: That’s just the tag. “v3” is updated when there’s a new release. See here how the sha is the same for v3 and v3.0.3

https://github.com/actions/setup-dotnet/tags

Shane32: Ignore dependabot

sungam3r: It means I don't know how versioning works in actions but I just did the same as we did in other repos.

Shane32:
```suggestion
      uses: actions/setup-dotnet@v3
```

.github/workflows/publish-preview.yml
```diff
@@ -21,10 +21,10 @@ jobs:
     steps:
       - name: Checkout source
         uses: actions/checkout@v3
-      - name: Setup .NET Core 6.0 SDK
-        uses: actions/setup-dotnet@v3
+      - name: Setup .NET SDK
+        uses: actions/setup-dotnet@v3.0.3
```
Shane32, line 25:
```suggestion
        uses: actions/setup-dotnet@v3
```

.github/workflows/publish-release.yml
```diff
@@ -27,10 +27,10 @@ jobs:
           version="${github_ref:10}"
           echo version=$version
           echo "version=$version" >> $GITHUB_ENV
-      - name: Setup .NET Core 6.0 SDK
-        uses: actions/setup-dotnet@v3
+      - name: Setup .NET SDK
+        uses: actions/setup-dotnet@v3.0.3
```
Shane32, line 31:
```suggestion
        uses: actions/setup-dotnet@v3
```

.github/workflows/test.yml
```diff
@@ -31,13 +31,14 @@ jobs:
     steps:
       - name: Checkout source
         uses: actions/checkout@v3
-      - name: Setup .NET Core SDKs
-        uses: actions/setup-dotnet@v3
+      - name: Setup .NET SDKs
+        uses: actions/setup-dotnet@v3.0.3
```
Shane32, line 35:
```suggestion
        uses: actions/setup-dotnet@v3
```

Shane32 (approved):

PR #295, opened 2023-04-16
Allow to parse value literals

sungam3r:
I need to parse values returned from introspection request, mostly default values of arguments. Our current design allows to parse only entire GraphQL document. 

Related project - https://github.com/sungam3r/graphql-introspection-model

### DIFF
--- a/src/GraphQLParser.ApiTests/GraphQLParser.approved.txt
+++ b/src/GraphQLParser.ApiTests/GraphQLParser.approved.txt
@@ -661,6 +661,7 @@ namespace GraphQLParser
     public static class Parser
     {
         public static GraphQLParser.AST.GraphQLDocument Parse(GraphQLParser.ROM source, GraphQLParser.ParserOptions options = default) { }
+        public static GraphQLParser.AST.GraphQLValue ParseValue(GraphQLParser.ROM source, GraphQLParser.ParserOptions options = default) { }
     }
     public struct ParserOptions
     {
--- a/src/GraphQLParser.Tests/ParserTests.cs
+++ b/src/GraphQLParser.Tests/ParserTests.cs
@@ -1,5 +1,6 @@
 using System.Collections;
 using System.Runtime.InteropServices;
+using GraphQLParser.Exceptions;
 
 namespace GraphQLParser.Tests;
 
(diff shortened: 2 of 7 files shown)


sungam3r:
@Shane32 I'm rewriting https://github.com/sungam3r/graphql-introspection-model now. I decided to completely remove my own `SDLBuilder` class that prints schema and switch to `SDLPrinter` from parser project. To be able to do this I wrote `ASTConverter` that converts my introspection representations into AST representations of GraphQL-Parser. Doing this I noticed that I need to properly handle all literals returned as default values or arguments. Fortunately all code I need is already here, but not as public API.

Shane32 (approved):

src/GraphQLParser.ApiTests/GraphQLParser.approved.txt
```diff
@@ -661,6 +661,7 @@ namespace GraphQLParser
     public static class Parser
     {
         public static GraphQLParser.AST.GraphQLDocument Parse(GraphQLParser.ROM source, GraphQLParser.ParserOptions options = default) { }
+        public static GraphQLParser.AST.GraphQLValue ParseValue(GraphQLParser.ROM source, GraphQLParser.ParserOptions options = default) { }
```
sungam3r, line 664: Just a note - it seems that we could make more methods public to allow parse concrete parts of document.

Shane32: Yeah I was thinking like "couldn't we just have Parse" return whatever was parsed?  Then cast it to whatever type was desired or something?  But the problem is that the parser wouldn't know the initial state -- if, for example, it was parsing a list of arguments or a list of directives.

We could perhaps add `Parse<T>` where T is a node type, and then a `switch typeof(T)` logic to bump into the correct code.

Shane32: Then you could write `Parse<GraphQLValue>` if you want that, or `Parse<List<GraphQLArgument>>` if you are parsing that, etc, with very little API changes and somewhat intuitive behavior.

Shane32: Not really a fan of a million different public `Parse...` methods.  Hopefully they are all exposed as protected methods or something so anyone CAN use them if needed.

sungam3r: This is a subject to discuss for v9.

sungam3r: #296

PR #300, opened 2023-04-17
Fix NRE when printing directive without locations

sungam3r:
Edge case in tests for https://github.com/sungam3r/graphql-introspection-model

### DIFF
--- a/src/GraphQLParser.Tests/Visitors/SDLPrinterTests.cs
+++ b/src/GraphQLParser.Tests/Visitors/SDLPrinterTests.cs
@@ -824,6 +824,46 @@ Line 4
     }
 
     [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public void Printer_Should_Print_DirectiveDefinition_Without_Locations(bool eachDirectiveLocationOnNewLine)
+    {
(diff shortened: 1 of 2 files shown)


Shane32 (approved):
The real problem is that we have `= null!;` as the property initializer, which we should not if we are exposing the class publicly.

sungam3r:
I can argue. `= null!` is done by following the GraphQL grammar - the AST node for locations is not optional. For optional nodes we do have `?` in their property type.

sungam3r:
I mean any property in AST node may be uninialized now if constructed manually. Parser throws exception if locations not set into parsed text.

Shane32:
Right but what I mean is that we should have a constructor which requires a value and remove the default constructor, in order for the NRT attributes to be accurate always. Then “= null!” is not required. Or, make the default constructors internal.

Certainly the NRT attributes are correct as-is to match the AST parsing.

Shane32:
Internally we can still do it the other way.  But the `public` constructor would have proper NRT attributes.  We can also allow for users to pass in `null!` if they like to the constructor, but they are then knowingly breaking NRT semantics.

sungam3r:
Understood.

PR #307, opened 2023-04-20
Make AST constructors parametrized

### LINKED ISSUE #302: Revise AST constructors
Internally we can still do it the other way.  But the `public` constructor would have proper NRT attributes.  We can also allow for users to pass in `null!` if they like to the constructor, but they are then knowingly breaking NRT semantics.

_Originally posted by @Shane32 in https://github.com/graphql-dotnet/parser/issues/300#issuecomment-1513459595_

sungam3r:
fixes #302

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,6 +1,6 @@
 test:
-  - src/GraphQLParser.ApiTests/**/*
   - src/GraphQLParser.Tests/**/*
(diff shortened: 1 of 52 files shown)


sungam3r:
This is the last PR of 3 I would like to merge to publish v9.

Shane32 (approved):

Shane32:
Looks great!

PR #316, opened 2023-04-21
Expose parameterless ctors for AST nodes

sungam3r:


Shane32:
Let's mark them as deprecated.  Ok?

sungam3r:
🤔  not needed anymore after https://github.com/graphql-dotnet/graphql-dotnet/pull/3599

PR #317, opened 2023-04-21
Add parametrized ctors to AST nodes used by GraphQL.NET v7

sungam3r:


Shane32 (approved):

sungam3r:
`v8` will be a long-lived branch like `v5` from core repo

Shane32:
It's so nice being able to easily push to any version on an as-needed basis.  Even now, if you delete the `v8` branch, it's already tagged, and can be brought back to an actual branch with a few clicks.

sungam3r:
Yep. And I want to note that with MinVer such behavior is preserved.

Shane32:
> Yep. And I want to note that with MinVer such behavior is preserved.

But it just isn't necessary.  Dependencies are security vulnerabilities.  They are dependabot nuisances. Why add another one?  So we don't have to change one line of code every few months (when we are already adding a new feature)?  If it's so much trouble, just set it to a high version number and leave it alone, like I've said repeatedly.

Perhaps if GitHub Actions did not produce a run number, or we did not set the version when publishing, it would be great.  That is not the case.  Nearly everything is already automated.

In short, I believe it will cause more trouble than it solves.

sungam3r:
It is not necessary. It is just convenient, simple because fits into our workflow. Just give it a try.

PR #319, opened 2023-04-26
Remove init-only setters for full .NET Standard / .NET Framework compatibility

Shane32:
Target Parser 10.x ??


sungam3r:
I would not. Moreover, I would not treat `init` with such caution. We have not yet written a single issue about problems with `init`. Nevertheless, I'm OK to merge this one as is.

sungam3r (approved):

PR #325, opened 2023-06-18
Add SDLSorter

Shane32:
This PR adds another visitor implementation to sort a SDL -- that is, to walk the tree and sort each list of nodes that are encountered.  Some node types, such as comments and literal arrays, are left in their original order.  Fortunately the implementation turned out to be quite simple to implement.  This implementation has a fixed sort order, generally by name ascending, but the string comparison used can be selected, with culture-invariant case-insensitive as default.  It can sort execution or definition documents.  Of course, execution documents will execute differently since the execution order is defined by the order of the fields in the selection set.  However, sorting definition documents will not cause any executional changes.  Some lists of nodes will have a primary sort based on what makes sense, with the secondary sort being by name:
1. A definition document will have the schema first, directives second, and types third
2. A schema definition will have the query first, mutation second, and subscription defined third
3. Executable documents will have queries first, mutations second, subscriptions third, and fragments fourth
4. Selection sets will have fields first, inline fragments second, and fragment spreads third

### DIFF
--- a/src/GraphQLParser.ApiTests/GraphQLParser.approved.txt
+++ b/src/GraphQLParser.ApiTests/GraphQLParser.approved.txt
@@ -763,7 +763,7 @@ namespace GraphQLParser.Visitors
         protected virtual System.Threading.Tasks.ValueTask VisitArgumentsAsync(GraphQLParser.AST.GraphQLArguments arguments, TContext context) { }
         protected virtual System.Threading.Tasks.ValueTask VisitArgumentsDefinitionAsync(GraphQLParser.AST.GraphQLArgumentsDefinition argumentsDefinition, TContext context) { }
         public virtual System.Threading.Tasks.ValueTask VisitAsync(GraphQLParser.AST.ASTNode? node, TContext context) { }
-        protected System.Threading.Tasks.ValueTask VisitAsync<T>(System.Collections.Generic.List<T>? nodes, TContext context)
+        protected virtual System.Threading.Tasks.ValueTask VisitAsync<T>(System.Collections.Generic.List<T>? nodes, TContext context)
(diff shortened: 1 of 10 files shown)


Shane32:
Note: sort algorithms were chosen based on benchmarks:

```
1000 items

|  Method |        Mean |     Error |    StdDev |      Median |   Gen0 |   Gen1 | Allocated |
|-------- |------------:|----------:|----------:|------------:|-------:|-------:|----------:|
| BuiltIn |    75.56 us |  2.070 us |  6.006 us |    77.90 us | 1.2207 |      - |   7.93 KB |
|    Linq |    99.10 us |  1.947 us |  2.164 us |    98.96 us | 5.7373 | 0.2441 |  35.48 KB |
|  Bubble | 3,287.07 us | 64.212 us | 71.371 us | 3,251.79 us |      - |      - |   7.87 KB |
|   Merge |   135.03 us |  2.669 us |  3.375 us |   133.80 us | 2.4414 |      - |   15.7 KB |

10 items

|  Method |     Mean |    Error |   StdDev |   Median |   Gen0 | Allocated |
|-------- |---------:|---------:|---------:|---------:|-------:|----------:|
| BuiltIn | 187.6 ns |  3.71 ns |  6.78 ns | 188.0 ns | 0.0317 |     200 B |
|    Linq | 412.4 ns |  8.12 ns | 10.84 ns | 415.2 ns | 0.1097 |     688 B |
|  Bubble | 310.1 ns |  6.22 ns |  8.09 ns | 312.5 ns | 0.0215 |     136 B |
|   Merge | 403.6 ns | 18.65 ns | 55.00 ns | 376.3 ns | 0.0381 |     240 B |

```

PR #326, opened 2023-06-18
Add PrintDeprecationReasons option

Shane32:
This feature mimics the similar feature available within GraphQL.NET's SchemaPrinter.  It defaults to true.  Disabling it simply omits the `reason` argument from `@deprecated` directives throughout the document.

Todo:
- add tests


Shane32:
I'm thinking maybe this should be handled by an AST filter within GraphQL.NET

sungam3r:
Delete branch?

PR #336, opened 2023-07-18
Add DefaultVisitorContext and NullVisitorContext

Shane32:
- Adds `NullVisitorContext` for use with synchronous visitors that do not hold context variables
- Adds `DefaultVisitorContext` for use with asynchronous visitors that do not hold context variables

### DIFF
--- a/src/GraphQLParser/Visitors/NullVisitorContext.cs
+++ b/src/GraphQLParser/Visitors/NullVisitorContext.cs
@@ -1,0 +1,12 @@
+namespace GraphQLParser.Visitors;
+
+/// <summary>
+/// An implementation of <see cref="IASTVisitorContext"/> that does nothing.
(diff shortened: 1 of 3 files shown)


src/GraphQLParser/Visitors/NullVisitorContext.cs
```diff
@@ -0,0 +1,12 @@
+namespace GraphQLParser.Visitors;
+
+/// <summary>
+/// An implementation of <see cref="IASTVisitorContext"/> that does nothing.
+/// Ideal for use in cases where the visitor runs synchronously, there is no context
+/// variables, and cancellation is not required.
+/// </summary>
+public struct NullVisitorContext : IASTVisitorContext
```
sungam3r, line 8: readonly struct?

PR #350, opened 2023-10-11
Bump xunit and related fixes

sungam3r:
I suggest to say goodbye for netcoreapp3.1 and net5.0 since the-latest-and-greatest xunit does not support these TFMs, see #344 and #345.

изображение

### DIFF
--- a/src/GraphQLParser.Tests/Visitors/ASTVisitorTests.cs
+++ b/src/GraphQLParser.Tests/Visitors/ASTVisitorTests.cs
@@ -23,17 +23,17 @@ public class ASTVisitorTests
     }
 
     [Fact]
-    public void ASTVisitor_Should_Throw_On_Unknown_Node()
+    public async Task ASTVisitor_Should_Throw_On_Unknown_Node()
     {
         var visitor = new ASTVisitor<Context>();
         var context = new Context();
 
-        var ex = Should.Throw<NotSupportedException>(() => visitor.VisitAsync(new MySuperNode(), context).GetAwaiter().GetResult());
+        var ex = await Should.ThrowAsync<NotSupportedException>(async () => await visitor.VisitAsync(new MySuperNode(), context));
         ex.Message.ShouldBe("Unknown node 'MySuperNode'.");
     }
 
     [Fact]
-    public void ASTVisitor_Should_Pass_CancellationToken()
+    public async Task ASTVisitor_Should_Pass_CancellationToken()
     {
         var document = "scalar JSON".Parse();
         var visitor = new MyVisitor();
@@ -41,7 +41,7 @@ public class ASTVisitorTests
         var context = new Context { CancellationToken = cts.Token };
         context.CancellationToken.ThrowIfCancellationRequested();
 
-        Should.Throw<OperationCanceledException>(() => visitor.VisitAsync(document, context).GetAwaiter().GetResult());
+        await Should.ThrowAsync<OperationCanceledException>(async () => await visitor.VisitAsync(document, context));
     }
 
     private sealed class MyVisitor : ASTVisitor<Context>
--- a/src/GraphQLParser.Tests/Visitors/CountVisitorTests.cs
+++ b/src/GraphQLParser.Tests/Visitors/CountVisitorTests.cs
@@ -38,7 +38,7 @@ field: Int }
 
         var document = text.Parse();
 
-        await visitor.VisitAsync(document, context).ConfigureAwait(false);
+        await visitor.VisitAsync(document, context);
         context.Count.ShouldBe(expectedCount);
         document.AllNestedCount().ShouldBe(expectedCount);
     }
@@ -55,7 +55,7 @@ field: Int }
 
         var document = text.Parse();
 
-        await visitor.VisitAsync(document, context).ConfigureAwait(false);
+        await visitor.VisitAsync(document, context);
         context.Count.ShouldBe(0);
     }
 }
--- a/src/GraphQLParser.Tests/Visitors/SDLPrinterFromParsedTextTests.cs
+++ b/src/GraphQLParser.Tests/Visitors/SDLPrinterFromParsedTextTests.cs
@@ -927,8 +927,8 @@ type Person {
         await printer.PrintAsync(document, writer);
         var renderedOriginal = writer.ToString();
 
-        var lines = renderedOriginal.Split(Environment.NewLine);
-        var renderedDescription = string.Join(Environment.NewLine, lines.SkipLast(1));
+        var lines = renderedOriginal.Split(new string[] { Environment.NewLine }, StringSplitOptions.None);
+        var renderedDescription = string.Join(Environment.NewLine, lines.Take(lines.Length - 1));
         renderedDescription = renderedDescription.Replace("\r\n", "\n");
         renderedDescription.ShouldBe(expected);
 
--- a/src/GraphQLParser.Tests/Visitors/SDLPrinterFromParsedTextTests.cs
+++ b/src/GraphQLParser.Tests/Visitors/SDLPrinterFromParsedTextTests.cs
@@ -867,7 +867,7 @@ type Person {
         var writer = new StringWriter();
         var document = text.Parse();
 
-        await printer.PrintAsync(document, writer).ConfigureAwait(false);
+        await printer.PrintAsync(document, writer);
         var actual = writer.ToString();
         actual.ShouldBe(expected, $"Test {number} failed");
 
@@ -924,7 +924,7 @@ type Person {
         var document = (input + " scalar a").Parse();
 
         var printer = new SDLPrinter();
-        await printer.PrintAsync(document, writer).ConfigureAwait(false);
+        await printer.PrintAsync(document, writer);
         var renderedOriginal = writer.ToString();
 
         var lines = renderedOriginal.Split(Environment.NewLine);
@@ -955,7 +955,7 @@ type Person {
         var document = query.Parse();
 
         var printer = new SDLPrinter();
-        await printer.PrintAsync(document, writer).ConfigureAwait(false);
+        await printer.PrintAsync(document, writer);
         var rendered = writer.ToString();
         rendered.ShouldBe(expected);
 
--- a/src/GraphQLParser/GraphQLParser.csproj
+++ b/src/GraphQLParser/GraphQLParser.csproj
@@ -4,17 +4,15 @@
     <Description>Library containing lexer and parser for GraphQL syntax</Description>
     <Product>GraphQL Parser</Product>
     <AssemblyTitle>GraphQL Parser for .NET</AssemblyTitle>
-    <TargetFrameworks>netstandard2.0;netstandard2.1;net6;net462</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;net6</TargetFrameworks>
     <PackageId>GraphQL-Parser</PackageId>
     <PackageTags>GraphQL;json;api;parser</PackageTags>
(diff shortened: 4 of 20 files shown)


src/GraphQLParser.Tests/Visitors/ASTVisitorTests.cs
```diff
@@ -23,25 +23,25 @@ public void ASTVisitor_Should_Handle_Null()
...
 
     [Fact]
-    public void ASTVisitor_Should_Throw_On_Unknown_Node()
+    public async Task ASTVisitor_Should_Throw_On_Unknown_Node()
     {
         var visitor = new ASTVisitor<Context>();
         var context = new Context();
 
-        var ex = Should.Throw<NotSupportedException>(() => visitor.VisitAsync(new MySuperNode(), context).GetAwaiter().GetResult());
+        var ex = await Should.ThrowAsync<NotSupportedException>(async () => await visitor.VisitAsync(new MySuperNode(), context));
```
sungam3r, line 31: https://xunit.net/xunit.analyzers/rules/xUnit1031

src/GraphQLParser.Tests/Visitors/CountVisitorTests.cs
```diff
@@ -38,7 +38,7 @@ public async Task CountVisitor_Should_Count_Nodes(string text, int expectedCount
 
         var document = text.Parse();
 
-        await visitor.VisitAsync(document, context).ConfigureAwait(false);
+        await visitor.VisitAsync(document, context);
```
sungam3r, line 41: https://xunit.net/xunit.analyzers/rules/xUnit1030

Shane32:
> I suggest to say goodbye for netcoreapp3.1 and net5.0 since the-latest-and-greatest xunit does not support these TFMs, see #344 and #345.
> 
> изображение

I still use older TFMs in various applications and it does not hurt us to test for them. If you want to bump xunit, just make the dependency conditional on the tfm and use the older version for 3.1/5.0.

src/GraphQLParser.Tests/Visitors/CountVisitorTests.cs
Shane32: Please note that we had to add ConfigureAwait calls to DataLoader tests to make them work properly. If we make similar changes in the main project we will need to be sure multithreaded tests work reliably.

sungam3r: I understand. It's a matter of balance. For these repo I think `.ConfigureAwait(false)` calls in tests are just noise.

Shane32:
Note that if we do drop .NET 3.1 testing, we implicitly drop .NET Standard 2.1 testing. 👎

Shane32:
And it seems that we don't test against the .NET Framework currently (a supported TFM), which means that we are not testing the .NET Standard 2.0 build. 👎

sungam3r:
True. Is it worth to?

Shane32:
If we have have different builds for .NET Standard 2.0/2.1 we should test against it.  It's not like it costs any dev time and CI testing is free.  It would be nice if we could test each of the 3 builds against .NET 6, but that would likely be quite a bit more complicated than simply testing .NET 4.8, .NET 3.1 and .NET 6.0.

I certainly have GraphQL applications that run on .NET Framework under a fully supported scenario, as I expect others do...

sungam3r:
OK, I'll deal with this tomorrow making dependency conditional.

Shane32:
> Arguably we could remove the .NET Standard 2.1 TFM from the codebase, using .NET Standard 2.0 as a fallback for older TFMs. I'm not prepared to do so at this time.

I'll also review this.  Perhaps there is little practical benefit to having a .NET Standard 2.1 TFM.

Shane32:
Looking at the codebase, there isn't any code that wouldn't be exercised if we tested on both .NET Framework 4.8 and .NET 6, omitting .NET 3.1.  Again, I'd probably just run it on 3.1 anyway, but if we do eliminate it, at least all the code paths get tested.

I wouldn't eliminate the .NET Standard 2.1 TFM.  For one thing, it eliminates some unnecessary dependencies.

Shane32:
We don't need to test against .NET 5.  It isn't a LTM release, and .NET 3.1 testing will cover the .NET Standard 2.1 binary.

src/GraphQLParser.Tests/Visitors/SDLPrinterFromParsedTextTests.cs
```diff
@@ -924,11 +924,11 @@ public async Task SDLPrinter_Should_Print_BlockStrings(int number, string input,
         var document = (input + " scalar a").Parse();
 
         var printer = new SDLPrinter();
-        await printer.PrintAsync(document, writer).ConfigureAwait(false);
+        await printer.PrintAsync(document, writer);
         var renderedOriginal = writer.ToString();
 
-        var lines = renderedOriginal.Split(Environment.NewLine);
-        var renderedDescription = string.Join(Environment.NewLine, lines.SkipLast(1));
+        var lines = renderedOriginal.Split(new string[] { Environment.NewLine }, StringSplitOptions.None);
```
sungam3r, line 930: I had to change signatures because of net462 TFM that lacks of previous signatures.

src/GraphQLParser/Visitors/SDLPrinterExtensions.cs
```diff
@@ -24,13 +24,28 @@ public static string Print(this SDLPrinter printer, ASTNode node)
...
+
     /// <summary>
     /// Prints the specified AST into the specified <see cref="MemoryStream"/> as a SDL document.
     /// If no encoding is specified, the document is written in UTF-8 format without a byte order mark.
     /// </summary>
     public static void Print(this SDLPrinter printer, ASTNode node, MemoryStream memoryStream, Encoding? encoding = null)
     {
-        using var streamWriter = new StreamWriter(memoryStream, encoding, -1 /* default */, true);
+        int bufferSize = -1;
+#if NET462
```
sungam3r, line 38: CoreCLR https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/IO/StreamWriter.cs#L94

NET Framework https://github.com/Microsoft/referencesource/blob/master/mscorlib/system/io/streamwriter.cs#L140

Actually this method **does not work at all** in NET Framework for now throwing ArgumentOutOfRangeException. Caller can pass not-null encoding, but buffer size is always negative. 😮

sungam3r: I did not touch signatures to not make breaking changes.

sungam3r: As option - add new method with bufferSize or change existing arguments in v10 to not bother with additional net462 TFM and all jumping around.

Shane32: Looking at the MS docs, we should not be passing `null` to `encoding`, nor should we be passing `-1` to `bufferSize`.  What we should learn here is that if we do not follow specs, then the code can fail.  For example, what would occur if someone tried to run this on Mono or Xamrin or some other .NET Standard implementation?  We cannot assume that `-1` and `null` would work there any more than it does on .NET Framework.  I would remove the ifdef and leave the code, so that no matter if it is .NET Standard or .NET 6/7/8/9/etc, it will adhere to the current specification, which lists these exceptions:

> ArgumentNullException : path or encoding is null.
> ArgumentOutOfRangeException : buffersize is less than or equal to zero.

We can view the source code of the constructor to determine what the default buffer size would otherwise be.

src/GraphQLParser/GraphQLParser.csproj
```diff
@@ -4,15 +4,17 @@
...
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Memory" Version="4.5.5" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
-    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
+    <Using Remove="System.Net.Http" Condition="'$(TargetFramework)' == 'net462'" />
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" Condition="'$(TargetFramework)' == 'net462'" />
+    <PackageReference Include="System.Memory" Version="4.5.5" Condition="'$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'net462'" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" Condition="'$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'net462'" />
```
Shane32, line 17: I would revert all changes here, so when we test against the .NET Framework, it uses the .NET Standard 2.0 TFM.  Note that if we do not revert these changes, we still do not have testing of the .NET Standard 2.0 build.  (We would then need to add testing against .NET Core 2.1 in order to test the .NET Standard 2.0 build.)

src/GraphQLParser/Visitors/SDLPrinterExtensions.cs
Shane32: > add new method with bufferSize or change existing arguments in v10

I would not do either at this time.  Our API dictates that null can be passed to encoding, which will use UTF8 with no byte order mark.  Fine.  It doesn't say what buffer size we will use.  Fine.  No breaking changes.  No need to make changes.  Our code simply doesn't work as we documented that it should.

Shane32: Sorry, I stand corrected; we are following the .NET 6 specs; but that overload does not exist for .NET Standard 2.0.  So for .NET 6 it says the arguments are optional when in .NET Standard 2.0/2.1 they are required.  So we can pass `-1` and `null` for .NET 6, but for .NET Standard we must pass correct values.  So let's just use `#if !NET6_0` or whatever rather than adding a TFM to the build process for .NET framework.

Shane32: https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.-ctor?view=net-6.0#system-io-streamreader-ctor(system-io-stream-system-text-encoding-system-boolean-system-int32-system-boolean)

https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.-ctor?view=netstandard-2.1#system-io-streamreader-ctor(system-io-stream-system-text-encoding-system-boolean-system-int32-system-boolean)

sungam3r: Fine. Done.

src/GraphQLParser/GraphQLParser.csproj
sungam3r: reverted

sungam3r:
Hah.
```
Test run for /home/runner/work/parser/parser/src/GraphQLParser.Tests/bin/Debug/net462/GraphQLParser.Tests.dll (.NETFramework,Version=v4.6.2)
Microsoft (R) Test Execution Command Line Tool Version 17.7.1 (x64)
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
[xUnit.net 00:00:02.58] GraphQLParser.Tests: Catastrophic failure: System.TypeLoadException: Could not load type of field 'Xunit.Runner.VisualStudio.VsExecutionSink:recorder' (4) due to: Could not load file or assembly 'Microsoft.VisualStudio.TestPlatform.ObjectModel, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies.
```

sungam3r:
9.2.1 ?

sungam3r:
I'm confused a bit. I just merged master into develop (actually fast-forwarded develop) and now master and develop both point to the same commit. What version to release - 8.4.1 or 9.2.1 ?

sungam3r:
I'm done for today. You may go on and make release if you wish.

Shane32:
Oh.  Sorry, we are using the `v8` branch for v8 and the `master` branch for v9

Shane32:
I changed the default branch to `v8` so we don't make this mistake again.

sungam3r:
Ok.

PR #355, opened 2023-10-13
Fix doubled whitespace for inline fragments

### LINKED ISSUE #352: There's an extra space here also, between `...` and `@skip`.
There's an extra space here also, between `...` and `@skip`.

_Originally posted by @Shane32 in https://github.com/graphql-dotnet/parser/pull/327#discussion_r1355256306_

sungam3r:
fixes #352

### DIFF
--- a/.github/codeql-config.yml
+++ b/.github/codeql-config.yml
@@ -1,0 +1,3 @@
+query-filters:
+  - exclude:
+      id: cs/linq/missed-where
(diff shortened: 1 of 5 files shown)


Shane32 (approved):

sungam3r:
@Shane32 See https://github.com/graphql-dotnet/parser/actions/runs/6503896922 . It seems that jumping with porting commits across branches may break something else.

sungam3r:
I copied `codeql-analysis.yml` from `master` into `v8`.

PR #359, opened 2023-10-16
Fix whitespace issues

sungam3r:


src/GraphQLParser.Tests/ParserTests.CustomAST.cs
```diff
@@ -79,6 +96,9 @@ public void Should_Parse_Directives()
         string text = "@my @your";
         var ast = text.Parse<GraphQLDirectives>().ShouldNotBeNull();
         ast.Items.Count.ShouldBe(2);
+
+        // TODO: fix leading space before first directive
```
sungam3r, line 100: ` @my @your` instead of `@my @your` now. It could be fixes as well checking parent node but for directives parent is always not null - it is `Directives` container, so actually grand parent is needed. Moreover, even after such check some tests failed (see `Printer_Should_Print_Pretty_If_Directives_Skipped`). The core problem is that some arbitrary nodes **can be** skipped so it's not obvious how to handle whitespaces at all. General approach may be to require each node to print their leading whitespace if it should do that. Downside - it will complicate all printing code. I do not want to do that just to handle very rare cases.

PR #361, opened 2023-10-16
Fix quad quote handling

Shane32:
Quad quotes within a block string was broken for parsing, and did not print well.

So this did not parse:

```gql
"""
Testing\""""123
"""
```

Which should parse as

```
Testing""""123
```

And when attempting to print it, it printed this, while valid, was not ideal:

```gql
"""
Testing\"\"""123
"""
```

Both are fixed in this PR.


sungam3r:
BlockStrings is not my strong skill. Approved.

sungam3r (approved):

Shane32:
I think I wrote some of this code originally but I could be wrong.

PR #371, opened 2023-11-13
Fix printing operation type

### LINKED ISSUE #370: Printing SDL drops ‘mutation’
When prettifying a mutation, the mutation keyword is removed from the output.

```c#
public static string Prettify(string query)
{
    var document = Parser.Parse(query);
    return new SDLPrinter().Print(document);
}

string query = Prettify("""
  mutation {
    bulkOperationCancel {
      bulkOperation {
        id
      }
    }
  }
  """);
```

Result:
```graphql
{
  bulkOperationCancel {
    bulkOperation {
      id
    }
  }
}
```

_Originally posted by @clement911 in https://github.com/graphql-dotnet/parser/issues/365#issuecomment-1807255141_

Shane32:
Fixes #370

### DIFF
--- a/src/GraphQLParser.Tests/Visitors/SDLPrinterTests.cs
+++ b/src/GraphQLParser.Tests/Visitors/SDLPrinterTests.cs
@@ -814,6 +814,17 @@ Line 4
     }
 
     [Theory]
+    [InlineData("{ field1 }", "{\n  field1\n}")]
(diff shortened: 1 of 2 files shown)


sungam3r:
@Shane32 I still can’t find enough time for normal project support. There are days when I go through hundreds of emails, but these are exceptions.

PR #391, opened 2024-08-18
Append newline to printed documents

Shane32:
I'd like this merged in for v9.4.0 in preparation for GraphQL.NET v8.0.  Note: GraphQL.NET v7.x uses the v8 parser.


src/GraphQLParser.Tests/Visitors/SDLSorterTests.SortsSchemaDefinition.approved.KitchenSink.graphql
```diff
@@ -139,4 +139,4 @@ extend type Foo {
   seven(argument: [String]): Type
 }
 
-extend type Foo @onType
\ No newline at end of file
+extend type Foo @onType
```
Shane32, line 142: This PR fixes these issues -- generated documents did not include a terminating newline before.

src/GraphQLParser/Visitors/SDLPrinter.cs
```diff
@@ -1097,13 +1097,15 @@ public SDLPrinter(SDLPrinterOptions options)
...
+    public virtual async ValueTask PrintAsync(ASTNode node, TextWriter writer, CancellationToken cancellationToken = default)
     {
         var context = new DefaultPrintContext(writer)
         {
             CancellationToken = cancellationToken,
         };
-        return VisitAsync(node, context);
+        await VisitAsync(node, context).ConfigureAwait(false);
+        if (!context.NewLinePrinted && node is GraphQLDocument)
+            await writer.WriteLineAsync().ConfigureAwait(false);
```
Shane32, line 1108: Changes were made only within `PrintAsync` and only for printed documents, not for individual nodes being printed, or where `VisitAsync` was called directly and the context could be examined to determine whether or not a newline had been printed.

src/GraphQLParser.Tests/Visitors/SDLSorterTests.SortsSchemaDefinition.approved.KitchenSink.graphql
Shane32, line 142: This PR fixes these issues -- generated documents did not include a terminating newline before.

gao-artur (approved):

src/GraphQLParser.Tests/Visitors/SDLSorterTests.SortsSchemaDefinition.approved.KitchenSink.graphql
sungam3r: Hi, Shane! For the last year or two I completely abandoned GraphQL projects and GitHub from the moment of changing my work. Now I slowly read the accumulated mail and watch what has been done during this time. My progress is small and I'm not sure that someday I will catch everything completely. When dependabot has updated the version of the parser in one of my projects, I found that all my tests failed. It brought me here. Why do you generally call it an _issue_? I did not specifically add a new line at the end of the document. Are you faced with a specific problem like the impossibility of a correct search using grep? Honestly, I myself 50/50 about newline usage. I just want to feel good, knowing that there is a good reason why it is better to have newline than without it.

sungam3r: Well, OK but I'm still 50/50 here. I updated my tests with trailing new line.

PR #398, opened 2024-10-19
Use auto properties and misc code cleanup

alexander-jesner-AP:
This PR adds a .NET8 compilation target to the main library and addresses performance related warnings that came up with the new build target.
Furthermore, this PR contains a few low hanging code cleanups .

### DIFF
--- a/src/GraphQLParser/Exceptions/GraphQLParserException.cs
+++ b/src/GraphQLParser/Exceptions/GraphQLParserException.cs
@@ -8,10 +8,12 @@ namespace GraphQLParser.Exceptions;
 /// </summary>
 public class GraphQLParserException : Exception
 {
-    /// <summary>
-    /// Error description.
-    /// </summary>
-    public string Description { get; private set; }
+    private static string[] NewlineSeparator = new string[] { "\n" };
+
+/// <summary>
+/// Error description.
+/// </summary>
+public string Description { get; private set; }
 
     /// <summary>
     /// Location of the symbol that caused the error.
@@ -49,7 +51,7 @@ public class GraphQLParserException : Exception
         int padLen = nextLineNum.Length;
         string[] lines = source
             .ToString()
-            .Split(new string[] { "\n" }, StringSplitOptions.None)
+            .Split(NewlineSeparator, StringSplitOptions.None)
             .Select(e => ReplaceWithUnicodeRepresentation(e))
             .ToArray();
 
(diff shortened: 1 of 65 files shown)


Shane32 (changes requested):
Thanks for the code cleanup! We won’t be adding a .NET 8 target unless there is some benefit. As it is now, the compiled code would be identical for .NET 6 or .NET 8. So adding a target would simply increase the size of the nuget download.

Shane32:
We could change the .net analyzers to use the .net 8 analyzers so these issues are displayed even when not targeting .net 8

alexander-jesner-AP:
I was under the impression, .NET8 generates slightly better MSIL. But I diffed the IL and you are right, a few attributes are no longer compiled in, but the rest is the same.

Shane32:
> I was under the impression, .NET8 generates slightly better MSIL. But I diffed the IL and you are right, a few attributes are no longer compiled in, but the rest is the same.

Likely enough the .NET 8 SDK does compile better than its predecessors, for any target. I’m not sure. I believe we are using the .NET 8 SDK for compilation.

src/GraphQLParser/Exceptions/GraphQLParserException.cs
```diff
@@ -8,6 +8,8 @@ namespace GraphQLParser.Exceptions;
 /// </summary>
 public class GraphQLParserException : Exception
 {
+    private static readonly char[] NewlineSeparator = ['\n'];
```
Shane32, line 11: Looks like this needs to be renamed to `_newlineSeparator` to match the naming rules enforced in this repo.

src/GraphQLParser/Exceptions/GraphQLParserException.cs
```diff
@@ -49,7 +51,7 @@ private static string HighlightSourceAtLocation(ReadOnlySpan<char> source, Locat
         int padLen = nextLineNum.Length;
         string[] lines = source
             .ToString()
-            .Split(new string[] { "\n" }, StringSplitOptions.None)
+            .Split(NewlineSeparator, StringSplitOptions.None)
```
Shane32, line 54: For .NET Standard 2.1 and .NET 6 TFMs we can use this overload:
https://learn.microsoft.com/en-us/dotnet/api/system.string.split?view=net-8.0#system-string-split(system-char-system-stringsplitoptions)

So we could do this:
```suggestion
#if NETSTANDARD2_1 || NET6_0_OR_GREATER
            .Split('\n', StringSplitOptions.None)
#else
            .Split(_newlineSeparator, StringSplitOptions.None)
#endif
```

src/GraphQLParser/Exceptions/GraphQLParserException.cs
Shane32: And with the below suggestion this would be:

```suggestion
#if NETSTANDARD2_0
    private static readonly char[] _newlineSeparator = ['\n'];
#endif
```

Shane32 (approved):

